PR #12, opened 2024-02-14
fix(crd): increasing the grace period for virtual machine shutdown

fl64:
## Description
 Set the termination grace period for virtual machine to 1 hour

The normal shutdown of a VM can take a long time. 1 hour is chosen as a compromise default setting.

## Checklist
- [ ] e2e tests passed.


yaroslavborbat (approved):

diafour (approved):
LGTM, 1  hour is a reasonable time to get lunch.

PR #19, opened 2024-03-01
feat(hooks): add generation of root certificate and certificates for module apps

yaroslavborbat:
## Description
Add a hook to generate RootCA and generate certificates from it

We will have a rootCA

after updating the virtualization module -> hook generates a rootCA and places it in secret virtualization-tls and cm virtualization-ca. All certificates will be regenerated from rootCA

## Checklist
- [ ] e2e tests passed.


templates/rootCA-secret.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: virtualization-ca
+  namespace: d8-{{ .Chart.Name }}
+  {{- include "helm_lib_module_labels" (list . (dict "app" "virtualization")) | nindent 2 }}
+type: kubernetes.io/tls
+data:
+  tls.crt: {{ .Values.virtualization.internal.admissionWebhookCert.crt }}
+  tls.key: {{ .Values.virtualization.internal.admissionWebhookCert.key }}
```
fl64, line 11: Why not `.Values.virtualization.internal.rootCA.{crt,key}` ?

fl64:
lgtm, but please fix PR message and description because it PR not only adding CA but also refactoring.

fl64 (approved):

PR #20, opened 2024-03-06
feat(api): moved virtualization api structs to separated go mod

Isteb4k:
## Description

Moved virtualization api to separated go module 

It is necessary to be able to import only api entities into other projects without the virtualization controller internals

New go module 

## Checklist
- [ ] e2e tests passed.


api/scripts/update-codegen.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+set -o errexit
+set -o nounset
+set -o pipefail
+
+function usage {
+  cat <<EOF
+Usage: $(basename "$0") <core/subresources/all>
+Example:
+   $(basename "$0") controller
```
yaroslavborbat, line 10:
```suggestion
   $(basename "$0") core
```

yaroslavborbat (approved):

images/virtualization-artifact/Taskfile.dist.yaml
```diff
@@ -682,6 +658,22 @@ tasks:
...
+    cmd: task api:generate:core
+
+  generate:subresources:
+    desc: "Regenerate code for subresources components"
+    cmd: task api:generate:subresources
+
+  ci:generate:
+    desc: "Run generations and check git diff to ensure all files are committed"
+    cmd: task api:ci:generate
+
```
diafour, line 676: I think these shortcuts are not needed. task list will show all api:NNNNN tasks.

images/virtualization-artifact/go.mod
```diff
@@ -3,20 +3,17 @@ module github.com/deckhouse/virtualization-controller
 go 1.21
 
 require (
+	github.com/deckhouse/virtualization/api v0.0.0-20240321085500-118bb9c256f6
```
diafour, line 6: Oops, `replace` is gone.

diafour (approved):
Just add replace.

PR #51, opened 2024-04-06
fix: fixed storage class and count processing

Isteb4k:
## Description
Fixed storage class and count processing

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/tests/performance/Taskfile.yaml
+++ b/tests/performance/Taskfile.yaml
@@ -5,9 +5,11 @@ version: "3"
 silent: true
 
 vars:
-  NAME_PREFIX: '{{ .NAME_PREFIX | default "performance" }}'
   COUNT: '{{ .COUNT | default "1" }}'
+  NAMESPACE: "{{ .NAMESPACE }}"
+  STORAGE_CLASS: "{{ .STORAGE_CLASS }}"
   RESOURCES: '{{ .RESOURCES | default "all" }}'
+  NAME_PREFIX: '{{ .NAME_PREFIX | default "performance" }}'
   RESOURCES_PREFIX: '{{ .RESOURCES_PREFIX | default "performance" }}'
 
 tasks:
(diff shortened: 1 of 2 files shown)


tests/performance/Taskfile.yaml
```diff
@@ -29,14 +31,33 @@ tasks:
...
+          NAMESPACE: "{{ .NAMESPACE }}"
+          STORAGE_CLASS: "{{ .STORAGE_CLASS }}"
+          NAME_PREFIX: "{{ .NAME_PREFIX }}"
+          RESOURCES_PREFIX: "{{ .RESOURCES_PREFIX }}"
+      - task: apply:vms
+        vars:
+          COUNT: "{{ .COUNT }}"
+          NAMESPACE: "{{ .NAMESPACE }}"
+          NAME_PREFIX: "{{ .NAME_PREFIX }}"
+          RESOURCES_PREFIX: "{{ .RESOURCES_PREFIX }}"
```
fl64, line 46: why duplicate variables if they are already globally declared?

tests/performance/Taskfile.yaml
```diff
@@ -29,14 +31,33 @@ tasks:
...
 
   destroy:all:
     desc: "Destroy disks and virtual machines in two steps (in two different releases)."
     cmds:
-      - task destroy:vms
-      - task destroy:disks
+      - task: destroy:vms
+        vars:
```
fl64, line 52: same question here

tests/performance/Taskfile.yaml
Isteb4k: When passing variables to the task (COUNT=5 task apply:all), in subtasks, the values of global variables remain unchanged; it's necessary to explicitly pass new values ​​for the variables.

tests/performance/Taskfile.yaml
Isteb4k: same answer here

tests/performance/Taskfile.yaml
fl64: :okay:

universal-itengineer (approved):

PR #53, opened 2024-04-08
fix: free some space on Github-hosted runners

diafour:
## Description

- Initial 20Gb of free space is not enough for building kubevirt and cdi.
- Remove preinstalled Docker images and some huge packages: android, codeql, swift, etc. to get extra ~20Gb of free space.
- Update arduino/setup-task to v2 to use Node 20.

Initial 20Gb of free space is not enough for building kubevirt and cdi.

Kubevirt and CDI components are successfully built after changing patches.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/.github/actions/remove-unwanted-software/action.yml
+++ b/.github/actions/remove-unwanted-software/action.yml
@@ -1,0 +1,134 @@
+# Usage:
+#
+# jobs:
+#   build_job:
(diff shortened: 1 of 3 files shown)


universal-itengineer (approved):
It looks good

PR #69, opened 2024-04-24
refactor(vmi-router): Use IP from our VirtualMachine

diafour:
## Description

- Use status.ipaddress instead of extracting ip from huge status in Kubevirt VM.
- Proper use of controller-runtime
- Drop kubevirt.io/client
- Update netlink, cilium, controller-runtime, k8 dependencies
- Add aliases for netlink methods to develop with comfort in non-linux OSes.

We dont'want to use kube-proxy with this component. Also, dependency hell was a tech debt.

## Checklist
- [ ] e2e tests passed.


fl64 (approved):
It looks ok, do you think `const table = 1490`, should I put it as an environment variable?

PR #83, opened 2024-05-06
fix(vm): add value of the guest os info

LopatinDmitr:
## Description

add simple wait for value of the guest os info with test period

solves the problem of missing guest os info field value
## Checklist
- [ ] e2e tests passed.


diafour (approved):

diafour:
Failed jobs are ok here.

PR #106, opened 2024-05-22
fix(api): add name suffix

Isteb4k:
## Description
Adjusting parameter names in accordance with the recommendations

We must follow the generally accepted [recommendations](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md)

## Checklist
- [ ] e2e tests passed.


fl64 (approved):

fl64:
I think the type of PR should be "fix" :)

PR #147, opened 2024-06-20
fix(vmbda): fix hotplug api call

Isteb4k:
## Description

1. Patched the virt-controller: it now communicates with virt-api via the virtualization-api.

Previously, during the hot plugging process, virt-controller communicated with virt-api through the KubeVirt API service. However, since this API service was removed, the hot plugging process stopped working.

2. The Protection service does not attempt to add a finalizer if the resource is in the process of being deleted (to avoid the error `No new finalizers can be added if the object is being deleted`).

## Checklist
- [ ] e2e tests passed.


images/virt-artifact/patches/017-fix-vmi-subresource-url.patch
```diff
@@ -0,0 +1,57 @@
+diff --git a/pkg/virt-operator/resource/generate/rbac/controller.go b/pkg/virt-operator/resource/generate/rbac/controller.go
```
yaroslavborbat, line 1: we can opt out of "VolumeRequests" in kvvm and using only in kvvmi. Then our subresource "addVolume" should mutate kvvm spec and delegate request to virt-api. if we do it , we won't need this patch.

yaroslavborbat (changes requested):

yaroslavborbat (approved):

PR #159, opened 2024-06-26
fix(vd): fix fake pvc resizing

Isteb4k:
## Description
- Took into account the case when the size of the PVC in the spec and status can differ (for example, when the spec is 1G, but the status shows 976563Ki).
- Added tests for ResizingHandler.
- Fixed watch for pvc (not a controller reference)
- Added a missed Resizing phase

## Checklist
- [ ] e2e tests passed.


LopatinDmitr (approved):

images/virtualization-artifact/pkg/controller/vd/internal/resizing.go
```diff
@@ -85,33 +86,58 @@ func (h ResizingHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (re
...
-			return reconcile.Result{}, nil
-		}
-
+	pvcSpecSize := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+	switch vdSpecSize.Cmp(pvcSpecSize) {
+	// Expected disk size is LESS THAN expected pvc size: no resize needed as resizing to a smaller size is not possible.
+	case -1:
 		condition.Status = metav1.ConditionFalse
 		condition.Reason = vdcondition.NotRequested
 		condition.Message = ""
```
diafour, line 95: I think we should say something to user about "LESS" situation, as it was with ErrTooSmallDiskSize error.

```suggestion
		condition.Message = fmt.Sprintf("The virtual disk size is too low. Should be >= %s", pvcSpecSize.String())
```

diafour (approved):

PR #168, opened 2024-06-27
fix(vm):  lifecycle vm

yaroslavborbat:
## Description
fix secret keys for provisioning. [userData, userdata]
add observedGeneration to VirtualMachine.Status
remove OwnerReference from pvc when DataVolume is removed

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/service/disk_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/disk_service.go
@@ -153,12 +153,14 @@ func (s DiskService) CleanUpSupplements(ctx context.Context, sup *supplements.Ge
 		if err != nil {
 			return false, err
 		}
-		pvc.ObjectMeta.OwnerReferences = slices.DeleteFunc(pvc.ObjectMeta.OwnerReferences, func(ref metav1.OwnerReference) bool {
-			return ref.Kind == "DataVolume"
-		})
-		err = s.client.Update(ctx, pvc)
-		if err != nil && !k8serrors.IsNotFound(err) {
-			return false, err
+		if pvc != nil {
+			pvc.ObjectMeta.OwnerReferences = slices.DeleteFunc(pvc.ObjectMeta.OwnerReferences, func(ref metav1.OwnerReference) bool {
+				return ref.Kind == "DataVolume"
+			})
+			err = s.client.Update(ctx, pvc)
+			if err != nil && !k8serrors.IsNotFound(err) {
+				return false, err
+			}
 		}
 	}
 
--- a/images/virtualization-artifact/pkg/controller/service/disk_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/disk_service.go
@@ -20,11 +20,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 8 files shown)


LopatinDmitr (dismissed):

images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
```diff
@@ -58,6 +58,22 @@ func (h *LifeCycleHandler) Handle(ctx context.Context, s state.VirtualMachineSta
...
 	changed := s.VirtualMachine().Changed()
 
+	defer func() {
+		if changed == nil {
+			return
+		}
+		if len(changed.Status.Conditions) == 0 {
+			changed.Status.ObservedGeneration = changed.GetGeneration()
+			return
+		}
```
diafour, line 68: Shouldn't we always update ObservedGeneration on every reconcile despite of changes?

images/virtualization-artifact/pkg/controller/vm/internal/provisioning.go
```diff
@@ -131,10 +131,18 @@ func (h *ProvisioningHandler) genConditionFromSecret(ctx context.Context, builde
...
+		if _, ok := secret.Data[key]; ok {
+			found = true
+			break
+		}
+	}
+	if !found {
 		builder.Status(metav1.ConditionFalse).
 			Reason2(vmcondition.ReasonProvisioningNotReady).
-			Message(fmt.Sprintf("Secret %q doesn't have key \"userdata\".", secretKey.String()))
+			Message(fmt.Sprintf("Secret %q doesn't have any keys such as %v.", keys, secretKey.String()))
```
diafour, line 145:
```suggestion
			Message(fmt.Sprintf("Secret %q should has one of data fields %v.", keys, secretKey.String()))
```

images/virtualization-artifact/pkg/controller/service/disk_service.go
```diff
@@ -148,6 +149,17 @@ func (s DiskService) CleanUpSupplements(ctx context.Context, sup *supplements.Ge
 		if err != nil && !k8serrors.IsNotFound(err) {
 			return false, err
 		}
+		pvc, err := s.GetPersistentVolumeClaim(ctx, sup)
```
Isteb4k, line 152: Check pvc == nil

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/service/disk_service.go
Isteb4k: by the way, the DataVolume can be deleted by the hands of users. In this case, the ownerRef will still be in the in PVC

images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
yaroslavborbat: I think we should update the ObservedGeneration when the ObservedGeneration of all conditions has been updated

Isteb4k (approved):

PR #201, opened 2024-07-09
fix(vm): fix vm-router panics when we delete a virtual machine. 

yaroslavborbat:
## Description
fix reconsile VM. refactor DeleteRoute. 

vm-router panics when we delete a virtual machine.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/vmi-router/controllers/vmirouter_reconciler.go
+++ b/images/vmi-router/controllers/vmirouter_reconciler.go
@@ -34,7 +34,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	virtv1alpha2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
+	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
 	"vmi-router/netlinkmanager"
 )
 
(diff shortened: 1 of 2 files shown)


images/vmi-router/controllers/vmirouter_reconciler.go
```diff
@@ -73,29 +74,24 @@ func (r *VMRouterReconciler) Reconcile(ctx context.Context, req reconcile.Reques
...
-	var isAbsent bool
-	err := r.client.Get(ctx, req.NamespacedName, &vm)
-	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			isAbsent = true
-		} else {
-			r.log.Error(err, fmt.Sprintf("fail to retrieve vm/%s", req.String()))
-			return reconcile.Result{}, err
-		}
+	vm := &virtv1alpha2.VirtualMachine{}
```
Isteb4k, line 77: Usually, if it's a pointer, we want to check it for nil. However, in the case of such a definition, vm == nil will always be false because it's a pointer to an empty resource, not nil.

images/vmi-router/netlinkmanager/manager.go
```diff
@@ -275,31 +278,23 @@ func getCiliumInternalIPAddress(node *ciliumv2.CiliumNode) string {
...
-	var vmIP string
-	if vm != nil {
-		vmIP = vm.Status.IPAddress
-	}
+func (m *Manager) DeleteRoute(vmKey types.NamespacedName, vmIP string) {
 	if vmIP == "" {
 		// Try to recover IP from the cache.
 		m.vmIPsLock.RLock()
-		vmIP = m.vmIPs[vmKey]
+		vmIP = m.vmIPs[vmKey.String()]
```
Isteb4k, line 285: It is safer to put elements in the map, also using the String() method from the types.NamespacedName, so as not to duplicate the logic of obtaining the key.

Isteb4k (changes requested):

Isteb4k (approved):

PR #206, opened 2024-07-10
feat(vm): add statisticHandler

yaroslavborbat:
## Description
add statisticHandler

We need to display statistics on the virtual machine

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/vm_reconciler.go
+++ b/images/virtualization-artifact/pkg/controller/vm/vm_reconciler.go
@@ -311,7 +311,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		r.logger.Info("Run handler", slog.String("name", h.Name()))
 		res, err := h.Handle(ctx, s)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


images/virtualization-artifact/pkg/controller/vm/vm_reconciler.go
```diff
@@ -308,10 +308,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	var handlerErr error
 
 	for _, h := range r.handlers {
-		r.logger.Debug("Run handler", slog.String("name", h.Name()))
+		r.logger.Info("Run handler", slog.String("name", h.Name()))
 		res, err := h.Handle(ctx, s)
 		if err != nil {
-			r.logger.Error("The handler failed with an error", slog.String("name", h.Name()), log.SlogErr(err))
```
Isteb4k, line 314: Why not error level? It looks like we want to know about it

Isteb4k (changes requested):

diafour (approved):

Isteb4k (approved):

PR #220, opened 2024-07-17
feat(vm): add pod handler

yaroslavborbat:
## Description
Add pod  handler

We should protect the pods because we need to receive a termination message.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/powerstate/shutdown_reason.go
+++ b/images/virtualization-artifact/pkg/controller/powerstate/shutdown_reason.go
@@ -24,27 +24,29 @@ import (
 	kvv1 "kubevirt.io/api/core/v1"
 )
 
+type GuestSignalReason string
+
 const (
 	// DefaultVMContainerName - a container name with virt-launcher, libvirt and qemu processes.
 	DefaultVMContainerName = "compute"
 
 	// GuestResetReason - a reboot command was issued from inside the VM.
-	GuestResetReason = "guest-reset"
+	GuestResetReason GuestSignalReason = "guest-reset"
 
 	// GuestShutdownReason - a poweroff command was issued from inside the VM.
-	GuestShutdownReason = "guest-shutdown"
+	GuestShutdownReason GuestSignalReason = "guest-shutdown"
 )
 
 // ShutdownReason returns a shutdown reason from the Completed Pod with VM:
 // - guest-reset — reboot was issued inside the VM
 // - guest-shutdown — poweroff was issued inside the VM
 // - empty string means VM is still Running or was exited without event.
-func ShutdownReason(kvvmi *kvv1.VirtualMachineInstance, kvPods *corev1.PodList) (bool, string) {
+func ShutdownReason(kvvmi *kvv1.VirtualMachineInstance, kvPods *corev1.PodList) ShutdownInfo {
 	if kvvmi == nil || kvvmi.Status.Phase != kvv1.Succeeded {
-		return false, ""
+		return ShutdownInfo{}
 	}
 	if kvPods == nil || len(kvPods.Items) == 0 {
-		return false, ""
+		return ShutdownInfo{}
 	}
 
 	// Sort Pods in descending order to operate on the most recent Pod.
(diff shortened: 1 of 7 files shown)


images/virtualization-artifact/pkg/controller/vm/internal/pod.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if isDeletion(current) {
+		objs := make([]client.Object, len(pods.Items))
+		for i, p := range pods.Items {
+			objs[i] = p.DeepCopy()
+		}
+		return reconcile.Result{}, h.protection.RemoveProtection(ctx, objs...)
```
diafour, line 60: I feel that we can pass pods.Items as-is without DeepCopy.

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	kvvmi, err := s.KVVMI(ctx)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	info := powerstate.ShutdownReason(kvvmi, pods)
+	if info.PodCompeted {
+		s.Shared(func(s *state.Shared) {
+			s.ShutdownInfo = info
+		})
```
diafour, line 70: Do we need mutex guard here? PodHandler and SyncKvvmHandler are always executed one-by-one, never in parallel.

images/virtualization-artifact/pkg/controller/powerstate/shutdown_reason.go
```diff
@@ -70,13 +72,19 @@ func ShutdownReason(kvvmi *kvv1.VirtualMachineInstance, kvPods *corev1.PodList)
...
 		}
 	}
 
-	return true, ""
+	return ShutdownInfo{PodCompeted: true, Pod: recentPod}
+}
+
+type ShutdownInfo struct {
+	Reason      GuestSignalReason
+	PodCompeted bool
```
diafour, line 88:
```suggestion
	PodCompleted bool
```
=)

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
yaroslavborbat: We should add a pointer to objs, but we cannot add &p.
```
pkg/controller/vm/internal/pod.go:58:15: exporting a pointer for the loop variable p (exportloopref)
                        objs[i] = &p

```
We need to copy the object. And we have two ways
1. 
```go
pod = p 
objs[i] = &pod
```
2.
```go
objs[i] = p.DeepCopy()
```
but the first way does not perform a  deep copy.

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
yaroslavborbat: So, you are right. Mb we don't need in mutex. this has been around for a long time and was added just in case for the future. 
Do you think we should remove this?

diafour: Yes, I think you can get/set s.ShutdownInfo without locks.

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
diafour: Oh, that's about array casting. May be create h.protection.RemoveProtectionFromPods to range []v1.Pod list and apply `h.protection.RemoveProtection` there without DeepCopy?

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
yaroslavborbat: fixed

images/virtualization-artifact/pkg/controller/vm/internal/pod.go
yaroslavborbat: fixed

diafour (approved):

PR #242, opened 2024-07-29
feat(vm-route-forge): impl route reconciliation

yaroslavborbat:
## Description
add route reconciliation on the host

We need to monitor the routes on the host, not just the VMs in the cluster.

## Checklist
- [ ] e2e tests passed.


images/vm-route-forge/README.md
```diff
@@ -1,4 +1,4 @@
-## vmi-router
+## vm-route-forge
 
 > **NOTE:** Not an accurate name, should be 'cilium-route-updater'.
```
diafour, line 4: We've discussed, cilium is not a good name. This note can be deleted now.

```suggestion
```

images/vm-route-forge/cmd/vm-route-forge/app/root.go
```diff
@@ -0,0 +1,200 @@
...
+
+	// Load configuration to connect to Kubernetes API Server.
+	kubeCfg, err := config.GetConfig()
+	if err != nil {
+		log.Error(err, "Failed to load Kubernetes config")
+		return err
+	}
+	kubeClient, err := kubernetes.NewForConfig(kubeCfg)
+	if err != nil {
+		log.Error(err, "Failed to het kubeclient")
```
diafour, line 123:
```suggestion
		log.Error(err, "Failed to create Kubernetes client")
```

images/vm-route-forge/cmd/vm-route-forge/app/root.go
```diff
@@ -0,0 +1,200 @@
...
+		netMgr,
+		sharedCache,
+		parsedCIDRs,
+		&log,
+	)
+	if err != nil {
+		log.Error(err, "Failed to create route controller")
+		return err
+	}
+	go routeCtrl.Run(ctx, 1)
```
diafour, line 178: What is magic "1"? Add const to describe what is it.

images/vm-route-forge/cmd/vm-route-forge/app/root.go
```diff
@@ -0,0 +1,200 @@
...
+
+	httpServer := server.NewServer(opts.ProbeAddr, kubeClient)
+	httpServer.InstallDefaultHandlers()
+	return httpServer.ListenAndServe()
+}
+
+func preRunSync(ctx context.Context, mgr *netlinkmanager.Manager) error {
```
diafour, line 185: Not a good name. You call **preRun**Sync from **run** method =)
Call it for example, startupSync.

images/vm-route-forge/internal/controller/route/route_controller.go
```diff
@@ -0,0 +1,292 @@
...
+func (c *Controller) Run(ctx context.Context, workers int) {
+	defer utilruntime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	c.log.Info("Starting namespace controller")
+	defer c.log.Info("Shutting down namespace controller")
+
+	if !cache.WaitForNamedCacheSync(controllerName, ctx.Done(), c.hasSynced) {
+		return
+	}
```
diafour, line 218: Isn't it a disaster? Should we log an error at this point?

images/vm-route-forge/internal/server/server.go
```diff
@@ -0,0 +1,86 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"net/http"
+
+	"github.com/emicklei/go-restful/v3"
```
diafour, line 22: Do we need a restful library for a simple healtz? =)))

templates/kubevirt/vm-route-forge/daemonset.yaml
```diff
@@ -31,34 +31,34 @@ spec:
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
```
diafour, line 33: There is no health probe in a DaemonSet manifest.

images/vm-route-forge/cmd/vm-route-forge/app/root.go
yaroslavborbat: renamed to startupSync

images/vm-route-forge/internal/controller/route/route_controller.go
yaroslavborbat: maybe, but I think it's not necessary because we need to wait for cache synchronization before running handlers. If the cache is not synchronized, the controller will stop. For example, in the namespace controller it works the same way [(link)](https://github.com/kubernetes/kubernetes/blob/0fc167103189a4d462c3cc7a17b360c3d998f4bf/pkg/controller/namespace/namespace_controller.go#L205). But i added log message

images/vm-route-forge/cmd/vm-route-forge/app/root.go
yaroslavborbat: fixed

images/vm-route-forge/internal/server/server.go
yaroslavborbat: So, maybe you're right, but I think in the future we'll want to add metric handlers, proofs, etc

templates/kubevirt/vm-route-forge/daemonset.yaml
yaroslavborbat: will be solved in another task

images/vm-route-forge/internal/controller/route/host.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	for ru := range ch {
+		if err := r.sync(ru); err != nil {
+			r.log.Error(err, "failed to sync route update")
+		}
+	}
+	return nil
+}
+
+func (r *HostRouteController) sync(ru netlink.RouteUpdate) error {
```
diafour, line 62: Please, add comment for this method which describe what route updates  are lead to VM reconcile and what updates we ignore.

Isteb4k (approved):

PR #246, opened 2024-07-31
chore(vmclass): fix docs and crds

fl64:
Update vmclass info in *.md and fixes for vm crds.

## Checklist
- [ ] e2e tests passed.


templates/vmclasses.yaml
```diff
@@ -0,0 +1,103 @@
+---
```
yaroslavborbat, line 1: we already have vmclasses [manifests](https://github.com/deckhouse/virtualization/blob/main/templates/virtualization-controller/vmclasses-default.yaml)

do we need to delete old manifests?

yaroslavborbat (changes requested):

templates/vmclasses.yaml
fl64: oops

yaroslavborbat (approved):

PR #267, opened 2024-08-09
fix(vmbda): write to condition message if disk is already attached to vm spec

Isteb4k:
## Description

1. For vmbda, added a message to the condition that indicates a disk conflict if disk is already added to the virtual machine spec
2. Fixed an issue that prevented adding a disk to the spec when the vmbda with that disk was deleted for the virtual machine.
3. Failed phase is no longer final for vmbda (it can transition to Pending).
4. New validator for VM: BlockDeviceSpecRefsValidator, which prevents duplicate spec.BlockDeviceRef entries.
5. VM validators have been moved to internal/validators.

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vmbda/internal/life_cycle.go
+++ b/images/virtualization-artifact/pkg/controller/vmbda/internal/life_cycle.go
@@ -93,11 +93,6 @@ func (h LifeCycleHandler) Handle(ctx context.Context, vmbda *virtv2.VirtualMachi
 		return reconcile.Result{}, nil
 	}
 
-	// Final phase: avoid processing VMDBA that had a conflict with another VMDBA previously.
-	if vmbda.Status.Phase == virtv2.BlockDeviceAttachmentPhaseFailed {
-		return reconcile.Result{}, nil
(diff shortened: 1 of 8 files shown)


images/virtualization-artifact/pkg/controller/vmbda/internal/life_cycle.go
```diff
@@ -94,9 +94,9 @@ func (h LifeCycleHandler) Handle(ctx context.Context, vmbda *virtv2.VirtualMachi
 	}
 
 	// Final phase: avoid processing VMDBA that had a conflict with another VMDBA previously.
-	if vmbda.Status.Phase == virtv2.BlockDeviceAttachmentPhaseFailed {
-		return reconcile.Result{}, nil
-	}
+	// if vmbda.Status.Phase == virtv2.BlockDeviceAttachmentPhaseFailed {
+	// 	return reconcile.Result{}, nil
```
LopatinDmitr, line 98: Need this comment block?

LopatinDmitr (approved):

PR #279, opened 2024-08-13
fix(vm): wait for virtual disk's target pvc to be created before start internal virtual machine

Isteb4k:
## Description

Wait for virtual disk's target pvc to be created before start internal virtual machine

## Checklist
- [x] e2e tests passed.


LopatinDmitr (approved):

images/virtualization-artifact/pkg/controller/vm/internal/block_device.go
```diff
@@ -237,60 +241,71 @@ func (h *BlockDeviceHandler) checkBlockDevicesSanity(vm *virtv2.VirtualMachine)
...
+			var canAttach bool
+
+			switch {
+			case len(vd.Status.AttachedToVirtualMachines) == 0:
+				canAttach = true
+			case len(vd.Status.AttachedToVirtualMachines) == 1:
+				if vd.Status.AttachedToVirtualMachines[0].Name != vm.GetName() {
+					canAttach = false
+					msg := fmt.Sprintf("unable to attach virtual disk %s because it is already attached to another virtual machine %s", vd.Name, vd.Status.AttachedToVirtualMachines[0].Name)
+					warnings = append(warnings, msg)
```
yaroslavborbat, line 282: I think we shouldn't return warnings. we create a warning event at this, and later we create an event again with these warnings

images/virtualization-artifact/pkg/controller/vm/internal/block_device.go
```diff
@@ -237,60 +241,71 @@ func (h *BlockDeviceHandler) checkBlockDevicesSanity(vm *virtv2.VirtualMachine)
...
+					h.recorder.Event(vm, corev1.EventTypeWarning, virtv2.ReasonVDAlreadyInUse, msg)
+				} else {
+					canAttach = true
 				}
+			default:
```
yaroslavborbat, line 287: I don't think we need default

images/virtualization-artifact/pkg/controller/vm/internal/block_device.go
```diff
@@ -237,60 +241,71 @@ func (h *BlockDeviceHandler) checkBlockDevicesSanity(vm *virtv2.VirtualMachine)
...
+			if !hasKey {
+				canStartKVVM = false
+				continue
+			}
 
-				if !attachReady || vd.Status.Target.PersistentVolumeClaim == "" {
-					continue
+			var canAttach bool
+
+			switch {
```
yaroslavborbat, line 275:
```
				canAttach := true
				for _, attached := range vd.Status.AttachedToVirtualMachines {
					if attached.Name != vm.GetName() {
						canAttach = false
						msg := fmt.Sprintf("VirtualDisk %s was attached to another virtual machine %q", vd.Name, attached.Name)
						log.Warn(msg)
						h.recorder.Event(vm, corev1.EventTypeWarning, virtv2.ReasonVDAlreadyInUse, msg)
					}
				}
```
maybe something like this? 

we check everything and at the same time create an event for each connected disk

yaroslavborbat (changes requested):

PR #291, opened 2024-08-15
chore(core): install necessary packages for redos when selinux enabled

universal-itengineer:
## Description

When selinux is enabled on RedOS, `selinux-policy` and `container-selinux` packages must be installed on the system for virtualization to work properly.

When selinux is enabled, virt-handler will not be able to run on nodes and creating virtual machines will not be possible

Virtualmachine can be started if selinux is enabled.

> Package `selinux-policy` require restart system, if not previously installed

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/templates/nodegroupconfiguration-redos.yaml
+++ b/templates/nodegroupconfiguration-redos.yaml
@@ -24,11 +24,11 @@ spec:
     # See the License for the specific language governing permissions and
     # limitations under the License.
 
-    bb-event-on 'bb-package-installed' 'post-install'
-    post-install() {
-      bb-log-info "Setting reboot flag due to selinux-policy was installed"
-      bb-flag-set reboot
-    }
+    # bb-event-on 'bb-package-installed' 'post-install'
+    # post-install() {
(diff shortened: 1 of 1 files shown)


templates/nodegroupconfiguration-redos.yaml
```diff
@@ -0,0 +1,75 @@
...
+    # distributed under the License is distributed on an "AS IS" BASIS,
+    # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    # See the License for the specific language governing permissions and
+    # limitations under the License.
+
+    # bb-event-on 'bb-package-installed' 'post-install'
+    # post-install() {
+    #   bb-log-info "Setting reboot flag due to selinux-policy was installed"
+    #   bb-flag-set reboot
+    # }
```
fl64, line 31: Why we need to keep these commented lines?

universal-itengineer: We don't. Fixed

hardcoretime (dismissed):

hardcoretime (approved):

PR #296, opened 2024-08-19
fix(vmip): fix bug with create VirtualMachineIPAddress in different namespace, when VirtualMahineIPAddressLease 'Released'

LopatinDmitr:
## Description

- Add requeue after 30s
- Fix message in recording event
- Add log for case, when create VirtualMachineIPAddress in different namespace

Fix bug with create VirtualMachineIPAddress in different namespace, when VirtualMahineIPAddressLease 'Released'

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vmip/internal/iplease_handler.go
+++ b/images/virtualization-artifact/pkg/controller/vmip/internal/iplease_handler.go
@@ -66,7 +66,7 @@ func (h IPLeaseHandler) Handle(ctx context.Context, state state.VMIPState) (reco
 	condition, _ := service.GetCondition(vmipcondition.BoundType, vmipStatus.Conditions)
 
 	switch {
-	case lease == nil && vmipStatus.Address != "" && condition.Reason != vmipcondition.VirtualMachineIPAddressLeaseAlready:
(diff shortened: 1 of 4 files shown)


images/virtualization-artifact/pkg/controller/vmip/internal/iplease_handler.go
```diff
@@ -63,9 +63,10 @@ func (h IPLeaseHandler) Handle(ctx context.Context, state state.VMIPState) (reco
 	if err != nil {
 		return reconcile.Result{}, err
 	}
+	condition, _ := service.GetCondition(vmipcondition.BoundType, vmipStatus.Conditions)
 
 	switch {
-	case lease == nil && vmipStatus.Address != "":
+	case lease == nil && vmipStatus.Address != "" && condition.Reason != vmipcondition.VirtualMachineIPAddressLeaseAlready:
```
Isteb4k, line 69: Rename reason `VirtualMachineIPAddressLeaseAlready`. It's not clear: already what?

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vmip/internal/iplease_handler.go
LopatinDmitr: Fixed

Isteb4k (approved):

PR #299, opened 2024-08-20
chore(core): hardcoded module name

universal-itengineer:
## Description

Hardcoded module name

We don't need to use this environment variable anymore (use thouse variable can affected module like tis [PR 9051](https://github.com/deckhouse/deckhouse/pull/9051)

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/hooks/lib/hooks/common.py
+++ b/hooks/lib/hooks/common.py
@@ -15,7 +15,6 @@
 
 PUBLIC_DOMAIN_PREFIX = "%PUBLIC_DOMAIN%://"
 CLUSTER_DOMAIN_PREFIX = "%CLUSTER_DOMAIN%://"
-MODULE_NAME = "virtualization"
 
 def cluster_domain_san(san: str) -> str:
     """
--- a/hooks/lib/hooks/common.py
+++ b/hooks/lib/hooks/common.py
@@ -15,7 +15,7 @@
 
 PUBLIC_DOMAIN_PREFIX = "%PUBLIC_DOMAIN%://"
 CLUSTER_DOMAIN_PREFIX = "%CLUSTER_DOMAIN%://"
-
+MODULE_NAME = "virtualization"
 
 def cluster_domain_san(san: str) -> str:
     """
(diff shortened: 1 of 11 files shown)


hooks/lib/hooks/common.py
```diff
@@ -15,7 +15,7 @@
 
 PUBLIC_DOMAIN_PREFIX = "%PUBLIC_DOMAIN%://"
 CLUSTER_DOMAIN_PREFIX = "%CLUSTER_DOMAIN%://"
-
+MODULE_NAME = "virtualization"
```
yaroslavborbat, line 18: this is a library. We don't have to hard code our module in this.
We already have this var https://github.com/deckhouse/virtualization/blob/683bb70318210e89755de237d813a9720911395d/hooks/common.py#L25

hooks/lib/module/module.py
```diff
@@ -39,26 +40,4 @@ def get_https_mode(module_name: str, values: dict) -> str:
...
-            #                                  or
-            # /deckhouse/modules/900-virtualization/hooks/hook_name.py
-            # {---------------------- file_path ----------------------}
-            # {-- MODULES_DIR --}{---{-------- regexp result --------}}
-            #                        virtualization/hooks/hook_name.py
-            #                        {-module-name-}{-----------------}
-
-            break
-    return module
+    return MODULE_NAME
```
yaroslavborbat, line 43: I think, that we must always pass the module name as an argument. This feature should just be removed

yaroslavborbat: should delete:
1)
https://github.com/deckhouse/virtualization/blob/683bb70318210e89755de237d813a9720911395d/hooks/lib/hooks/hook.py#L25
2)
https://github.com/deckhouse/virtualization/blob/683bb70318210e89755de237d813a9720911395d/hooks/lib/hooks/hook.py#L43

yaroslavborbat (changes requested):

yaroslavborbat (dismissed):

PR #312, opened 2024-08-27
fix(vd): ensure last transition time for conditions

Isteb4k:
## Description

1. Ensure for all controllers that LastTransitionTime reflects the last time the condition transitioned from one status to another (not just any condition update), in accordance with the Kubernetes documentation.

2. Get rid of the deprecated Reason method and NewConditionBuilder function. Use the renamed Reason2 method and NewConditionBuilder2 function instead.

3. The conditions.Manager is now deprecated. It is recommended to use conditions.SetCondition instead.

No motivation was found to use conditions.Manager over directly updating conditions. Since conditions.Manager is deprecated, it is recommended to use conditions.SetCondition instead.

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/conditions/manager.go
+++ b/images/virtualization-artifact/pkg/controller/conditions/manager.go
@@ -19,14 +19,17 @@ package conditions
 import (
 	"slices"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Deprecated: use direct SetCondition instead.
 type Manager struct {
 	conds      []metav1.Condition
 	indexConds map[string]int
 }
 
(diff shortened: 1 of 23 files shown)


images/virtualization-artifact/pkg/controller/conditions/manager.go
```diff
@@ -19,14 +19,17 @@ package conditions
 import (
 	"slices"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Deprecated: use direct SetCondition instead.
```
yaroslavborbat, line 26: Maybe don't mark it as deprecated but remove it

yaroslavborbat (approved):

LopatinDmitr (approved):

PR #320, opened 2024-08-29
refactor(kubevirt): virsh error workaround for some kernels

diafour:
## Description

- Add override for getpeercon method from libselinux. Overridden method change EINVAL errno to ENOPROTOOPT, to libvirt feels happy.

Some implementations of getsockopt() syscall returns EINVAL errno for unsupported
valopt argument. getsockopt() is used by getpeercon() method from libselinux.
libvirt expects ENOPROTOOPT or ENOSYS from getpeercon().

Kubevirt's virt-handler works as expected.

## Checklist
- [x] e2e tests passed.


universal-itengineer (approved):
The tests were carried out by me manually - everything works correctly

universal-itengineer (approved):

PR #328, opened 2024-09-02
feat(cdi): remove service-monitor

yaroslavborbat:
## Description
Removed the creation of a service monitor from the cdi-operator.

Service Monitor is already being deployed using the virtualization module. We don't need 2 replicas

## Checklist
- [ ] e2e tests passed.


diafour:
I see no reasons to add unused code deletion to the patch. These deletions increase patch size and hide significant changes.

diafour (approved):

Isteb4k (approved):

PR #351, opened 2024-09-09
feat(vd): override virtualdisk's pvc parameters via StorageClass annotations

yaroslavborbat:
## Description
override pvc parameters using annotations in StorageClass

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/service/disk_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/disk_service.go
@@ -76,13 +76,36 @@ func (s DiskService) Start(
 	dvBuilder.SetDataSource(source)
 	dvBuilder.SetOwnerRef(obj, obj.GroupVersionKind())
 
+	sc, err := s.GetStorageClass(ctx, storageClass)
+	if err != nil {
+		return err
+	}
+
+	var (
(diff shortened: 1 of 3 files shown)


images/virtualization-artifact/pkg/controller/service/disk_service.go
```diff
@@ -296,7 +319,42 @@ func getAccessModeMax(modes []corev1.PersistentVolumeAccessMode) corev1.Persiste
 	return m
 }
 
-func (s DiskService) parseVolumeMode(status cdiv1.StorageProfileStatus) StorageCapabilities {
+func (s DiskService) parseVolumeMode(sc *storev1.StorageClass) (corev1.PersistentVolumeMode, bool) {
+	if sc == nil {
+		return "", false
+	}
+	switch sc.GetAnnotations()[common.AnnVirtualDiskVolumeMode] {
```
fl64, line 326: There may not be annotations at all

yaroslavborbat: This works correctly. If the annotation does not exist, the default case will be processed.

fl64 (approved):

PR #364, opened 2024-09-12
chore(core): add comments to prevent accidental adding status subresource

diafour:
## Description

- Add comments to prevent accidental adding status subresource in InternalVirtualizationCDI and InternalVirtualizationVirtualMachineInstance.

Some resources for KubeVirt and CDI store status field as "main" fields, not as a subresource. Set status to subresource will break update operation for these resources.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/crds/embedded/cdi.yaml
+++ b/crds/embedded/cdi.yaml
@@ -2320,3 +2320,8 @@ spec:
       served: true
       storage: true
       subresources: {}
+# IMPORTANT! Do not add status subresource, as it breaks the status update
+# in CDI components.
+# Delete this precaution when future CDI updates add status subresource.
+#      subresources:
+#        status: {}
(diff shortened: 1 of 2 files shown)


diafour:
Status field of VMI should not be a subresource. kubevirt updates status field as a part of "main" payload.

crds/embedded/cdi.yaml
```diff
@@ -1337,4 +1337,6 @@ spec:
           type: object
       served: true
       storage: true
-      subresources: {}
```
fl64, line 1340: Is it normal if we delete this fragment?

diafour: Fixed

fl64 (approved):

PR #366, opened 2024-09-12
chore(lint): add doc-ru linter

fl64:
## Description
- add doc-ru linter to check the differences between crd and the RU version of crd
- add task `lint` to run all linters

We need to make sure that all changes are reflected in all language versions of crd

A tool that allows you to display the difference in crd

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -87,6 +87,16 @@ tasks:
           go run tools/addlicense/{main,variables,msg,utils}.go -directory ./
         {{end}}
 
+  lint:doc-ru:
+    desc: "Check the correspondence between description fields in the original crd and the Russian language version"
+    cmds:
+    - |
+        docker run \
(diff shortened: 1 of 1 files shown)


Taskfile.yaml
```diff
@@ -87,6 +87,25 @@ tasks:
...
 
+  lint:
+    cmds:
+      - task: lint:doc-ru
+      - task: lint:prettier:yaml
+      - task: virtualization-controller:dvcr:_ensure:golangci-lint
+      - task: virtualization-controller:dvcr:lint
+      - task: virtualization-controller:dvcr:lint:go
+      - task: virtualization-controller:lint
+      - task: virtualization-controller:lint:go
```
diafour, line 98: Seems only 2 tasks are enough. (lint task executes lint:go task)

```suggestion
      - task: virtualization-controller:dvcr:lint
      - task: virtualization-controller:lint
```

diafour (approved):

PR #391, opened 2024-09-19
chore(crds): reformat yaml to kubectl style, ignore prettier

diafour:
- Ignore prettier for embedded CRDs
- No changes in CRD structure, only formatting.

Updating embedded CRDs lead to a huge diff that is hard to review. Use kubectl output and ignore prettier to get smaller diffs in subsequent PRs.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,6 +5,7 @@
 **/werf*.yml
 **/templates
 **/ISSUE_TEMPLATE
+crds/embedded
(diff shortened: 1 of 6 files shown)


diafour:
Mark as draft, not works as expected.

diafour:
There is no way to reduce diff between 1.0.0 and 1.3.1 CRDs after huge reformat in https://github.com/kubevirt/kubevirt/pull/11482

PR #413, opened 2024-10-01
refactor(ci): get rid of bazel build

universal-itengineer:
## Description

Switch to building kubevirt components on go and werf instead of building in bazel and werf

We need to simplify the build, make it more predictable and possible in closed environments

All pods with prefix `virt-` in `d8-virtualization` will be restarted
All VM `'firmware'` will be updated by vm migration 

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/libguestfs/werf.inc.yaml
+++ b/images/libguestfs/werf.inc.yaml
@@ -23,10 +23,12 @@ shell:
   - mkdir -p /usr/local/lib/guestfs/
   - curl -L {{ $libguestfsApplianceTar }} | tar -xJ -C /usr/local/lib/guestfs/
   - apt-get remove curl --yes
+  - apt-get clean
+  - rm --recursive --force /var/lib/apt/lists/ftp.altlinux.org* /var/cache/apt/*.bin
   install:
   # Install main packages, update GPG keys and vendor IDs list.
   - |
-    apt-get install --yes \
+    apt-get update && apt-get install --yes \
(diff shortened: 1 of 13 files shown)


images/libguestfs/werf.inc.yaml
```diff
@@ -1,25 +1,32 @@
...
   - .version
-  - entrypoint.sh
-  to: /
   before: setup
 shell:
+  beforeInstall:
+  - apt-get update && apt-get install --yes curl
+  - mkdir -p /usr/local/lib/guestfs/
+  - curl -L {{ $libguestfsApplianceTar }} | tar -xJ -C /usr/local/lib/guestfs/
+  - apt-get remove curl --yes
```
diafour, line 25: Append `apt clean` and `rm -rf` commands to reduce stage image size.

universal-itengineer: Fixed

images/virt-launcher/werf.inc.yaml
```diff
@@ -114,6 +117,7 @@ shell:
     seabios \
     libffi8 \
     swtpm-tools \
+    glibc \
```
diafour, line 120: Seems redundant. glibc should be installed in the base image.

universal-itengineer: You're right.
Fixed.

PR #418, opened 2024-10-03
feat(vm): add failure message to condition if migration is failed

yaroslavborbat:
## Description
add failure message to condition if migration is failed

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
@@ -127,22 +127,31 @@ func (h *LifeCycleHandler) syncMigrationState(vm *virtv2.VirtualMachine, kvvm *v
 	} else {
 		vm.Status.MigrationState = h.wrapMigrationState(kvvmi.Status.MigrationState)
 	}
+	setCondition := newSetCondition(vm)
 
-	//nolint:staticcheck
-	mgr := conditions.NewManager(vm.Status.Conditions)
 	cbMigrating := conditions.NewConditionBuilder(vmcondition.TypeMigrating).Generation(vm.GetGeneration())
-	if vm.Status.MigrationState != nil &&
+
+	switch {
+	case vm.Status.MigrationState != nil &&
 		vm.Status.MigrationState.StartTimestamp != nil &&
-		vm.Status.MigrationState.EndTimestamp == nil {
-		mgr.Update(cbMigrating.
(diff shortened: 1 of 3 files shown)


images/virtualization-artifact/pkg/controller/vm/internal/util.go
```diff
@@ -159,3 +159,9 @@ func isInternalVirtualMachineError(phase virtv1.VirtualMachinePrintableStatus) b
 func podFinal(pod corev1.Pod) bool {
 	return pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed
 }
+
+func newSetCondition(vm *virtv2.VirtualMachine) func(c conditions.Conder) {
```
Isteb4k, line 163: Don't see the reason to wrap here
Just use `conditions.SetCondition(c, &vm.Status.Conditions)` and keep it simple

images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
```diff
@@ -127,83 +127,80 @@ func (h *LifeCycleHandler) syncMigrationState(vm *virtv2.VirtualMachine, kvvm *v
...
+
+	case kvvmi != nil && kvvmi.Status.MigrationState != nil &&
+		kvvmi.Status.MigrationState.EndTimestamp != nil &&
+		kvvmi.Status.MigrationState.Failed:
+
+		msg := kvvmi.Status.MigrationState.FailureReason
+		cbMigrating.Status(metav1.ConditionFalse).
+			Reason(vmcondition.ReasonLastMigrationFinishedWithError).
+			Message(msg)
+		setCondition(cbMigrating)
```
Isteb4k, line 150: You don't do return, is it expected? 
Looks like this error can be overridden below

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vm/internal/util.go
yaroslavborbat: ok

images/virtualization-artifact/pkg/controller/vm/internal/lifecycle.go
yaroslavborbat: this is expected. the syncMigrationState method updates 2 conditions, so I don't return

Isteb4k (approved):

PR #431, opened 2024-10-09
fix(vm): fix the virtual machine matching with the virtual machine class during validation when no sizing policy is provided

eofff:
## Description
Fix Virtual Machine validation of matching Virtual Machine Class requirements when no Sizing Policy not provided.

Virtual Machines was failing validation when no sizing policy provided. That's incorrect behavior.

Correct validation process)

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/service/size_policy_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/size_policy_service.go
@@ -24,7 +24,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/resource"
 
-	"github.com/deckhouse/virtualization/api/core/v1alpha2"
+	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


images/virtualization-artifact/pkg/controller/service/size_policy_service.go
```diff
@@ -39,7 +39,12 @@ func NewSizePolicyService() *SizePolicyService {
 	return &SizePolicyService{}
 }
 
-func (s *SizePolicyService) CheckVMMatchedSizePolicy(vm *v1alpha2.VirtualMachine, vmClass *v1alpha2.VirtualMachineClass) error {
+func (s *SizePolicyService) CheckVMMatchedSizePolicy(vm *virtv2.VirtualMachine, vmClass *virtv2.VirtualMachineClass) error {
+	// check if no sizing policy requirements setted
```
Isteb4k, line 43: setted -> set

Isteb4k: Check if no sizing policy requirements are set.

Isteb4k (changes requested):

Isteb4k (approved):

PR #446, opened 2024-10-15
refactor(ci): switch from merge commit to head sha commit in builds

universal-itengineer:
## Description

Switch from Merge commit to HEAD SHA to prevent rebuilding images that have not been modified in the current branch

Improve CI/CD

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/.github/workflows/dev_module_build.yml
+++ b/.github/workflows/dev_module_build.yml
@@ -32,6 +32,7 @@ on:
     branches:
       - main
       - pre-alpha
+      # - refactor/ci/pullrequest-head-sha
 
 defaults:
   run:
@@ -121,6 +122,8 @@ jobs:
           version: 3.37.2
 
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Run lint virtualization-controller
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/dev_module_build.yml
```diff
@@ -26,7 +26,7 @@ env:
 on:
   workflow_dispatch:
   pull_request:
-    types: [opened, edited, synchronize]
+    types: [opened, synchronize]
```
diafour, line 29:
```suggestion
    types: [opened, reopened, synchronize]
```

I don't know if we need "reopened" here. It is rare we reopen PRs.

.github/workflows/dev_module_build.yml
```diff
@@ -120,6 +120,8 @@ jobs:
           version: 3.37.2
 
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
diafour, line 124: "github.event.pull_request" is empty for the "workflow_dispatch:" trigger.

.github/workflows/dev_module_build.yml
universal-itengineer: I think you're right, we need this option.

diafour (dismissed):

diafour (dismissed):

diafour (approved):

PR #451, opened 2024-10-15
refactor(kube-api-rewriter): rename kube-api-proxy to kube-api-rewriter

diafour:
## Description

- Rename kube-api-proxy directory in images directory.
- Refactor Helm templates
  - One templates for the sidecar and one for the Pod spec patch instead of 3 different sidecars and 3 different patches.
  - Move templates related to kube-api-rewriter into kube-api-rewriter directory.
  - Move up vm-route-forge directory.
  - Remove _helpers.tpl in cdi and kubevirt.
  - Split kube-api-rewriter helpers into 3 files, add documentation.
  - Unification: use KUBECONFIG env instead of args for the virt-api component.
  - Fix indent: move pprof container port into proxy container for virt-contoller,  virt-handler, exportproxy and cdi-deployment.

- "Rewriter" is more specific than "proxy".
- Also, "kube-api-proxy" name is somewhat misleading, because there is the "kube-proxy" component in the Deckhouse.
- Unified template for kube-api-rewriter is less error-prone.

Everything works as before, no errors in logs.

## Checklist
- [ ] e2e tests passed.


images/kube-api-rewriter/local/Dockerfile
```diff
@@ -1,5 +1,8 @@
+# Build kube-api-rewriter for local development purposes.
+# Note: it is not a part of the production build!
+
 # Go builder.
-FROM golang:1.21-alpine3.19 AS builder
+FROM golang:1.22-alpine3.19 AS builder
```
universal-itengineer, line 5: We need to use the version of the image containing a specific version of the go language to avoid downloading another version
```suggestion
FROM golang:1.22.7-alpine3.19 AS builder
```

images/kube-api-rewriter/local/Dockerfile
```diff
@@ -9,10 +12,10 @@ RUN go mod download
 ADD . /app
 
 RUN GOOS=linux \
-    go build -o proxy ./cmd/kube-api-proxy
+    go build -o kube-api-rewriter ./cmd/kube-api-rewriter
 
 # Go builder.
-FROM golang:1.21-alpine3.19 AS builder-test-controller
+FROM golang:1.22-alpine3.19 AS builder-test-controller
```
universal-itengineer, line 18: We need to use the version of the image containing a specific version of the go language to avoid downloading another version
```suggestion
FROM golang:1.22.7-alpine3.19 AS builder-test-controller
```

universal-itengineer (dismissed):

images/kube-api-rewriter/local/test-controller/go.mod
```diff
@@ -1,4 +1,4 @@
-module kube-api-proxy-test-controller
+module kube-api-rewriter-test-controller
 
 go 1.21
```
universal-itengineer, line 3: Maybe we should use version 1.22.7 here?
```suggestion
go 1.22.7
```

yaroslavborbat (approved):

yaroslavborbat (approved):

universal-itengineer (approved):

PR #453, opened 2024-10-16
refactor(api): unified deletion message when removing cleanup finalizer

diafour:
## Description
- Add log.Info before removing cleanup finalizer.
- Change confusing message "Delete %SOME_RESOURCE%" to "Deletion observed".

"Delete %SOME_RESOURCE%" message leads to a wrong assumption that controller deletes that %SOME_RESOURCE%.

## Checklist
- [ ] e2e tests passed.


diafour:
Info level is debatable, but I think deletion is worth mentioning in logs.

Isteb4k (approved):

PR #456, opened 2024-10-17
refactor(ci): run helm-templates validation on related changes only

diafour:
## Description

Add "dorny/paths-filter" action to detect changes in helm templates related files.

No need to run helm templates validation for any change in the repo.

helm templates validation is skipped for changes in Go sourcres.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/.github/workflows/dev_validation.yaml
+++ b/.github/workflows/dev_validation.yaml
@@ -30,6 +30,26 @@ defaults:
     shell: bash
 
 jobs:
+  paths_filter:
+    runs-on: ubuntu-latest
+    name: Filter changed paths
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/dev_validation.yaml
```diff
@@ -30,6 +30,27 @@ defaults:
...
+        with:
+          filters: |
+            helm_templates:
+              - 'crds/**'
+              - 'charts/**'
+              - 'tools/kubeconform/**'
+              - 'templates/**'
+              - .helmignore
+              - Chart.yaml
+              - Taskfile.yaml
```
universal-itengineer, line 52: I believe that Taskfile.yaml has mistakenly fallen into this list.
```suggestion
```

universal-itengineer (approved):

PR #475, opened 2024-10-28
fix(kube-api-rewriter): fix discovery and ValidatingAdmissionPolicy rewrite

yaroslavborbat:
## Description

- Fix array construction for discovery rewrite in v1.30.
- Add rewriting for ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding.

Make kube-api-rewriter compatible with Kubernetes 1.29+

Module works in Kubernetes 1.29+

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/kube-api-proxy/Taskfile.dist.yaml
+++ b/images/kube-api-proxy/Taskfile.dist.yaml
@@ -8,7 +8,7 @@ includes:
     optional: true
 
 vars:
-  DevImage: "localhost:5000/$USER/kube-api-proxy:latest"
+  DevImage: "${DevImage:-localhost:5000/$USER/kube-api-proxy:latest}"
 
 tasks:
   default:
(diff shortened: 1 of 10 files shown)


images/kube-api-proxy/Taskfile.dist.yaml
```diff
@@ -8,7 +8,7 @@ includes:
     optional: true
 
 vars:
-  DevImage: "localhost:5000/$USER/kube-api-proxy:latest"
+  DevImage: "${DevImage:-localhost:5000/$USER/kube-api-proxy:latest}"
```
LopatinDmitr, line 11: Are you sure there's no address error here?

diafour: localhost:5000 is for the local development.

diafour (dismissed):

diafour (approved):

PR #478, opened 2024-10-29
test(ci): add e2e github workflow

hardcoretime:
## Description

E2E tests will run every night.
Also it fix race condition in IP address assigning operation of `custom-ip` testcase.

It improves development experience.

We can find some of bugs before releases.
## Checklist
- [ ] e2e tests passed.


.github/workflows/e2e_main.yaml
```diff
@@ -0,0 +1,71 @@
...
+# limitations under the License.
+
+name: e2e
+
+env:
+  CI_COMMIT_REF_NAME: ${{ github.ref_name }}
+  GO_VERSION: "1.22.7"
+  GINKGO_VERSION: "2.20.0"
+
+on:
```
universal-itengineer, line 22:
```suggestion
on:
  workflow_dispatch:
```
I think we may need to run this workflow manually.

universal-itengineer (approved):

PR #485, opened 2024-10-30
fix(api): fix empty type conditions state

eofff:
## Description
Fix empty type conditions state

Fix situations when added conditions with empty type. This situations may causes controller logic errors.

Working controllers

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/conditions/condition.go
+++ b/images/virtualization-artifact/pkg/controller/conditions/condition.go
@@ -33,7 +33,10 @@ func GetConditionByType(condType string, conds []metav1.Condition) (metav1.Condi
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


images/virtualization-artifact/pkg/controller/conditions/condition.go
```diff
@@ -33,7 +33,10 @@ func GetConditionByType(condType string, conds []metav1.Condition) (metav1.Condi
 		}
 	}
 
-	return metav1.Condition{}, false
+	return metav1.Condition{
```
Isteb4k, line 36: Not found is not ConditionUnknown, it's empty

Isteb4k (changes requested):

PR #487, opened 2024-10-30
feat(api): improve crd conditions

yaroslavborbat:
## Description
Make all crds have the same conditions

We have different specifications of conditions.
There is no validation of conditions in crds.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/crds/clustervirtualimages.yaml
+++ b/crds/clustervirtualimages.yaml
@@ -193,36 +193,78 @@ spec:
               type: object
               properties:
                 conditions:
-                  description: |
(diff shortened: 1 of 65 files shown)


crds/clustervirtualimages.yaml
fl64, line 1: Please also update the RU documentation.
To check fields with missing description you can use `task lint:doc-ru`.

Isteb4k (dismissed):

Isteb4k (approved):

Isteb4k (approved):

PR #504, opened 2024-11-06
feat(cvi): namespace validation for vi/vd ObjectRef

danilrwx:
## Description
Currently, when specifying ObjectRef in ClusterVirtualImage to namespace scoped VirtualDisk and VirtualImage resources lack validation for the presence of namespace, its minimum and maximum lengths

If namespace is absent, ObjectRef cannot be found and its presence and correctness must be validated in the manifest.

Adding validation of namespace presence, its minimum and maximum length, but only for ObjectRefs like VirtualImage and VirtualDisk.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/crds/clustervirtualimages.yaml
+++ b/crds/clustervirtualimages.yaml
@@ -158,7 +158,7 @@ spec:
                         - rule: "self.kind == 'VirtualImage' || self.kind == 'VirtualDisk' ? has(self.__namespace__) : true"
                           message: The namespace is required for VirtualDisk and VirtualImage
                         - rule: "self.kind == 'VirtualImage' || self.kind == 'VirtualDisk' ? has(self.__namespace__) && size(self.__namespace__) > 0 : true"
-                          message: The namespace name must not be empty
+                          message: The namespace is required for VirtualDisk and VirtualImage
                         - rule: "self.kind == 'VirtualImage' || self.kind == 'VirtualDisk' ? has(self.__namespace__) && size(self.__namespace__) < 64 : true"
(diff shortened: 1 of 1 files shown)


crds/clustervirtualimages.yaml
```diff
@@ -154,6 +154,13 @@ spec:
                       description: |
                         Use an existing `VirtualImage`, `ClusterVirtualImage` or `VirtualDisk` to create an image.
                       required: ["kind", "name"]
+                      x-kubernetes-validations:
+                        - rule: "self.kind == 'VirtualImage' || self.kind == 'VirtualDisk' ? has(self.__namespace__) : true"
+                          message: The namespace is required for VirtualDisk and VirtualImage
```
yaroslavborbat, line 159: It looks like we don't need this rule.

crds/clustervirtualimages.yaml
```diff
@@ -154,6 +154,13 @@ spec:
...
+                        - rule: "self.kind == 'VirtualImage' || self.kind == 'VirtualDisk' ? has(self.__namespace__) && size(self.__namespace__) > 0 : true"
+                          message: The namespace name must not be empty
```
yaroslavborbat, line 161:
```suggestion
                          message: The namespace is required for VirtualDisk and VirtualImage
```

yaroslavborbat (changes requested):

crds/clustervirtualimages.yaml
danilrwx: removed the first validation

yaroslavborbat (approved):

PR #512, opened 2024-11-10
test(api): add resource deletion step in e2e test cases

hardcoretime:
- exclude namespace in resource deletion step;
- refactor `apply` and `delete` kubectl commands;
- handle errors in clean up phase;
- add resource deletion step in each test case.

## Description

When test case is completed it's resources should be deleted.

To catch errors when virtualization resources cannot be deleted correctly.

All resources in E2E tests should be deleted correctly in each test case. 
## Checklist
- [x] e2e tests passed.

### DIFF
--- a/tests/e2e/kubectl/kubectl.go
+++ b/tests/e2e/kubectl/kubectl.go
@@ -29,24 +29,27 @@ import (
 )
 
 const (
-	Cmd              = "kubectl"
-	ShortTimeout     = 10 * time.Second
-	MediumTimeout    = 30 * time.Second
-	LongTimeout      = 60 * time.Second
-	ExecExtraTimeout = 20 * time.Second
+	Cmd                             = "kubectl"
+	ShortTimeout                    = 10 * time.Second
+	MediumTimeout                   = 30 * time.Second
+	LongTimeout                     = 60 * time.Second
+	ExecExtraTimeout                = 20 * time.Second
+	Filename         FilenameOption = "--filename"
+	Kustomize        FilenameOption = "--kustomize"
 )
 
-type Resource string
+type (
+	Resource       string
+	FilenameOption string
+)
 
 type Kubectl interface {
-	Apply(filepath string, opts ApplyOptions) *executor.CMDResult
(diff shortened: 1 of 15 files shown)


tests/e2e/kubectl/kubectl.go
```diff
@@ -56,10 +59,17 @@ type Kubectl interface {
...
+// FilenameOption:
+//
+//	kubectl.Filename  // --filename
+//	kubectl.Kustomize // --kustomize
 type ApplyOptions struct {
-	Namespace string
-	Output    string
-	Force     bool
+	Filename       []string
+	FilenameOption FilenameOption
```
yaroslavborbat, line 68: We already have Filename. Why is it need?

tests/e2e/sizing_policy_test.go
```diff
@@ -239,4 +246,22 @@ var _ = Describe("Sizing policy", ginkgoutil.CommonE2ETestDecorators(), func() {
...
+				FilenameOption: kc.Kustomize,
+			})
+			Expect(res.Error()).NotTo(HaveOccurred(), "cmd: %s\nstderr: %s", res.GetCmd(), res.StdErr())
+			res = kubectl.Delete(kc.DeleteOptions{
+				Filename:       []string{newVmClassFilePath},
+				FilenameOption: kc.Filename,
+			})
+			Expect(res.Error()).NotTo(HaveOccurred(), "cmd: %s\nstderr: %s", res.GetCmd(), res.StdErr())
+		})
+	})
```
yaroslavborbat, line 266: Can we take it out for reuse?

yaroslavborbat (changes requested):

tests/e2e/kubectl/kubectl.go
hardcoretime: FilenameOption is used for --kustomize/--filename options.

tests/e2e/sizing_policy_test.go
hardcoretime: Good idea.

yaroslavborbat (approved):

PR #515, opened 2024-11-11
test(api): refactor e2e label and annotation test case

hardcoretime:
## Description

- adapt existing label and annotation test case to updated API
- fix typo and improve test cases contexts

It adapts `Label and Annotation` test case to updated Virtualization API and this test case does not be skipped in E2E workflow. 

`Label and Annotation` will be checked again in E2E workflow.
## Checklist
- [x] e2e tests passed.


tests/e2e/vm_label_annotation_test.go
```diff
@@ -18,263 +18,295 @@ package e2e
...
-		return cmdResult
+func RemoveLabel(resource kc.Resource, labels map[string]string, names ...string) error {
+	formattedLabels := make([]string, 0, len(labels))
+	for k := range labels {
+		formattedLabels = append(formattedLabels, fmt.Sprintf("%s-", k))
+	}
+	rawResources := strings.Join(names, " ")
+	rawLabels := strings.Join(formattedLabels, "")
+	subCmd := fmt.Sprintf("label %s %s --namespace %s %s", resource, rawResources, conf.Namespace, rawLabels)
+	res := kubectl.RawCommand(subCmd, kc.MediumTimeout)
```
eofff, line 55: Why you import with alias kc, but use full form?

diafour:
> Label and Annotation test case will be added to E2E tests workflow.

Is it "Label and Annotation test case is now a part of E2E tests workflow again."

And, shouldn't we unlock this test for `task run` also?

tests/e2e/testdata/vm-label-annotation/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: testcases
+namePrefix: pr-number-or-commit-hash-
+resources:
+  - ns.yaml
+  - vi
+  - overlays/vm-labal-annotation
```
diafour, line 8: Typo in the directory name: vm-label-annotation

tests/e2e/vm_label_annotation_test.go
```diff
@@ -18,263 +18,295 @@ package e2e
...
-			ApplyFromFile(vmManifest)
-			WaitVmStatus(vm.Name, VMStatusRunning)
-		})
-
-		AfterAll(func() {
-			By("Delete manifest")
-			kubectl.Delete(kc.DeleteOptions{
-				Filename:       []string{vmManifest},
-				FilenameOption: kc.Filename,
+	Context("When resources are applied:", func() {
```
diafour, line 105: ":" is not necessary in context description

tests/e2e/vm_label_annotation_test.go
```diff
@@ -18,263 +18,295 @@ package e2e
...
-		Describe(fmt.Sprintf("Add annotation %s=%s", annotationName, annotationValue), func() {
-			It("Annotated", func() {
-				subCMD := fmt.Sprintf("-n %s annotate vm %s %s=%s", conf.Namespace, vm.Name, annotationName,
-					annotationValue)
-				res := kubectl.RawCommand(subCMD, ShortWaitDuration)
-				Expect(res.Error()).NotTo(HaveOccurred())
-			})
-		})
+	Describe("Annotating", func() {
+		Context(fmt.Sprintf("When virtual machines in %s phase:", PhaseRunning), func() {
```
diafour, line 225: There is only one Context in this Describe. It seems more convenient to merge them into single Context:
```
Context(fmt.Sprintf("Annotate VirtualMachines in %s phase", PhaseRunning), func() ...
```

tests/e2e/vm_label_annotation_test.go
```diff
@@ -18,263 +18,295 @@ package e2e
...
-				Expect(res.Error()).NotTo(HaveOccurred(), "failed to get KVVMI %s.\n%s", vm.Name, res.StdErr())
-				Expect(res.StdOut()).To(BeEmpty())
-			})
-			It("POD virtlauncher", func() {
-				pod := getPodName(vm.Name)
-				res := getRecourseAnnotation(kc.ResourcePod, pod)
-				Expect(res.Error()).NotTo(HaveOccurred(), "failed to get pod %s.\n%s", vm.Name, res.StdErr())
-				Expect(res.StdOut()).To(BeEmpty())
+	Context("When test is complited:", func() {
+		It("tries to delete used resources", func() {
```
diafour, line 301:
```suggestion
	Context("When test is completed", func() {
		It("should delete all created resources", func() {
```

diafour, line 301: Also,  cleanup operation seems strange to be a part of the test suite. Shouldn't it belongs to the test setup actions, something like AfterAll?

eofff (approved):

tests/e2e/vm_label_annotation_test.go
hardcoretime: I put it to wrapper function with description. We need this context to check that all resources in test cases can be deleted correctly.

hardcoretime:
> > Label and Annotation test case will be added to E2E tests workflow.
> 
> Is it "Label and Annotation test case is now a part of E2E tests workflow again."?
> 
> And, shouldn't we unlock this test for `task run` also?

Yes, we unlock this test in this PR.

eofff (approved):

PR #526, opened 2024-11-15
feat(vd): block resizing VirtualDisk if storage class not ready

eofff:
## Description
Do not allow resizing operation of `VirtualDisk` if its `StorageClass` is not ready.

## Checklist
- [x] e2e tests passed.


yaroslavborbat (commented):
correct me if I'm wrong. The condition StorageClassReady means that the storage class exists. And we can't increase the size, but why? What problem does this PR solve?

yaroslavborbat (approved):

yaroslavborbat (approved):

PR #556, opened 2024-12-05
feat(vd): set tolerations for provisioners

Isteb4k:
## Description

**Problem scenario:**
Create local virtual disks (available only on the node where they were created) and a virtual machine on the masters, set tolerations and node selectors, but nothing starts because there are no tolerations for the disk provisioners.

**Solution:**
When creating virtual disks, the tolerations for the disk provisioners are inherited from the virtual machine and virtual machine class.

## Checklist
- [ ] e2e tests passed.


images/virtualization-artifact/pkg/controller/vd/internal/source/sources.go
```diff
@@ -234,6 +238,161 @@ func setPhaseConditionForPVCProvisioningDisk(
...
+
+		var isChanged bool
+		isChanged, err = provisioner.IsNodePlacementChanged(nodePlacement, pod)
+		if err != nil {
+			setPhaseConditionToFailed(cb, &vd.Status.Phase, fmt.Errorf("unexpected error: %w", err))
+			return err
+		}
+
+		if isChanged {
+			// log.Info("The node placement has changed for importer pod: recreate the provisioner")
```
LopatinDmitr, line 274: Need delete this?

LopatinDmitr (approved):

LopatinDmitr (approved):

PR #557, opened 2024-12-05
feat(vm): add specchanges for vmclass

yaroslavborbat:
## Description
- add annotation to save vmclass spec that was used to generate current kvvm.

We need to detect changes in vmclass to set status that VM needs a restart to apply destructive changes from the vmclass spec.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
@@ -103,12 +103,31 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 	}
 
 	// 1. Set RestartAwaitingChanges.
-	var lastAppliedSpec *virtv2.VirtualMachineSpec
-	var changes vmchange.SpecChanges
+	var (
+		lastAppliedSpec *virtv2.VirtualMachineSpec
+		changes         vmchange.SpecChanges
+		allChanges      vmchange.SpecChanges
+	)
 	if kvvm != nil {
 		lastAppliedSpec = h.loadLastAppliedSpec(current, kvvm)
 		lastClassAppliedSpec := h.loadClassLastAppliedSpec(class, kvvm)
-		changes = h.detectSpecChanges(ctx, kvvm, &current.Spec, lastAppliedSpec, &class.Spec, lastClassAppliedSpec)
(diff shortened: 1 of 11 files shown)


Isteb4k (dismissed):

images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
```diff
@@ -164,6 +180,16 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 			Status(metav1.ConditionTrue).
 			Reason(vmcondition.ReasonRestartAwaitingChangesExist).
 			Message("Waiting for the user to restart in order to apply the configuration changes.")
+	case classChanged:
+		h.recorder.Event(current, corev1.EventTypeNormal, virtv2.ReasonErrRestartAwaitingChanges, "The virtual machine configuration successfully synced")
```
diafour, line 184:
```suggestion
		h.recorder.Event(current, corev1.EventTypeNormal, virtv2.ReasonErrRestartAwaitingChanges, "Restart required to propogate changes from the vmclass spec")
```

images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
```diff
@@ -97,13 +97,29 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
...
+		allChanges      vmchange.SpecChanges
+		classChanged    bool
+	)
 	if kvvm != nil {
 		lastAppliedSpec = h.loadLastAppliedSpec(current, kvvm)
+		lastClassAppliedSpec := h.loadClassLastAppliedSpec(class, kvvm)
 		changes = h.detectSpecChanges(ctx, kvvm, &current.Spec, lastAppliedSpec)
+		if !changes.IsEmpty() {
+			allChanges.Add(changes.GetAll()...)
+		}
```
diafour, line 118: I don't get it, why this copy "allChanges" is needed?

yaroslavborbat: lost `allChanges.Add(classChanges.GetAll()...)`
allChanges contains vmChanges + classChanges.
This is necessary to preserve the old logic for applying the VM configuration

diafour: Oh, it make sense now.

diafour (approved):

PR #561, opened 2024-12-09
fix(vm): unfreeze fs after snapshot

danilrwx:
## Description
Unfreeze vm fs after snapshot

Currently, with multiple disks and their snapshots, we do not unfreeze the fs

fs unfreeze after all snapshots done

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/service/snapshot_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/snapshot_service.go
@@ -105,7 +105,9 @@ func (s *SnapshotService) CanUnfreeze(ctx context.Context, vdSnapshotName string
 
 		_, ok := vdByName[vdSnapshot.Spec.VirtualDiskName]
 		if ok {
-			return false, nil
+			if vdSnapshot.Status.Phase != virtv2.VirtualDiskSnapshotPhaseReady && vdSnapshot.Status.Phase != virtv2.VirtualDiskSnapshotPhaseFailed {
+				return false, nil
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


images/virtualization-artifact/pkg/controller/vmsnapshot/internal/life_cycle.go
```diff
@@ -72,6 +72,15 @@ func (h LifeCycleHandler) Handle(ctx context.Context, vmSnapshot *virtv2.Virtual
 	switch vmSnapshot.Status.Phase {
 	case "":
 		vmSnapshot.Status.Phase = virtv2.VirtualMachineSnapshotPhasePending
+	case virtv2.VirtualMachineSnapshotPhaseFailed:
```
Isteb4k, line 75: Add the same case for vd snapshot controller

images/virtualization-artifact/pkg/controller/service/snapshot_service.go
```diff
@@ -119,7 +121,9 @@ func (s *SnapshotService) CanUnfreeze(ctx context.Context, vdSnapshotName string
 
 	for _, vmSnapshot := range vmSnapshots.Items {
 		if vmSnapshot.Spec.VirtualMachineName == vm.Name {
-			return false, nil
+			if vmSnapshot.Status.Phase != virtv2.VirtualMachineSnapshotPhaseReady && vmSnapshot.Status.Phase != virtv2.VirtualMachineSnapshotPhaseFailed {
```
Isteb4k, line 124: Looks like **&& != Terminating** is also needed here

images/virtualization-artifact/pkg/controller/service/snapshot_service.go
```diff
@@ -105,7 +105,9 @@ func (s *SnapshotService) CanUnfreeze(ctx context.Context, vdSnapshotName string
 
 		_, ok := vdByName[vdSnapshot.Spec.VirtualDiskName]
 		if ok {
-			return false, nil
+			if vdSnapshot.Status.Phase != virtv2.VirtualDiskSnapshotPhaseReady && vdSnapshot.Status.Phase != virtv2.VirtualDiskSnapshotPhaseFailed {
```
Isteb4k, line 108: Looks like && != Terminating is also needed here

Isteb4k (changes requested):

Isteb4k (approved):

PR #562, opened 2024-12-09
feat(vm): Add Events about power state changes

diafour:
## Description

- Implement EventRecorderLogger to record Event and write it log.

We need meaningful Events for powerstate changes.

## Checklist
- [ ] e2e tests passed.


images/virtualization-artifact/pkg/eventrecord/eventrecorderlogger.go
```diff
@@ -0,0 +1,123 @@
...
+	recorder := e.recorderProducer.GetEventRecorderFor(e.recorderKey(reason))
+	recorder.AnnotatedEventf(object, annotations, eventtype, reason, messageFmt, args...)
+}
+
+func (e *EventRecorderLoggerImpl) logf(involved client.Object, eventtype, reason, messageFmt string, args ...interface{}) {
+	if e.logger == nil {
+		return
+	}
+	e.logger.Info(
+		fmt.Sprintf(messageFmt, args...),
```
yaroslavborbat, line 108: Can we have a problem with this? If the fmt string and args do not match.

yaroslavborbat (dismissed):

images/virtualization-artifact/pkg/eventrecord/eventrecorderlogger.go
diafour: At least Goland can catch these problems. Didn't check golangci-lint though.
image

Isteb4k (changes requested):
Rename branch

Isteb4k (dismissed):

Isteb4k (approved):

PR #564, opened 2024-12-10
feat(core, kubevirt): add patch for hotplug container-disk

yaroslavborbat:
## Description
Add patch for hotplug container-disk

## `container-disk`

The `container-disk` is a program written in C used within KubeVirt to facilitate the mounting of container-based disk images into virtual machines. Its core function is to start up and create a UNIX socket within a specific directory. The program terminates when the socket is removed or upon receiving a `SIGTERM` signal.

#### Initialization

- A sidecar container, running the `container-disk` image, is created alongside the `virt-launcher` pod.
- An init-container in the `virt-launcher` pod copies the `container-disk` program to a shared `emptyDir` volume. This setup allows the sidecar to execute the program.

#### Socket Creation

- The `container-disk` program creates a socket in the `emptyDir` volume.
- This shared volume allows the `virt-handler` to locate the socket on the host machine at:  
  `/var/lib/kubelet/pods/.../volumes/kubernetes.io~empty-dir/`.

#### Socket Detection and Mounting

- Upon detecting the socket, `virt-handler` identifies it as a `container-disk` volume and retrieves its parent mount point.
- For a container runtime like `containerd`, the mount point resolves to the root filesystem of the pulled image, typically at:  
  `/run/containerd/io.containerd.runtime.v2.task/k8s.io/<uid>/rootfs/`.
- The disk image must be located at `disk/disk.img` within this filesystem and is mounted into the VM.

---

## HotPlug in KubeVirt

The HotPlug mechanism allows dynamic attachment of PVCs and `container-disk` volumes to a running VM by leveraging a separate `hotplug` pod.

### HotPlug Pod Setup

- A `hotplug` pod is created with the target PVCs mounted into an `emptyDir` volume under the `/hp` directory.
- The `container-disk` program runs in the `hotplug` pod to create the necessary sockets for these volumes.

### Volume Detection and Mounting

- The `virt-handler` locates the sockets on the host system at:  
  `/var/lib/kubelet/pods/<uid-hotplug-pod>/volumes/empty-dir/hp-disks/...`.
- For block devices, `virt-handler` creates a block device on the VM using `mknodat`.
- For file systems, the volume is mounted as a file.

### Unmounting

- The unmount process is identical to that of `hotplug PVCs`.
- The `emptyDir` resources are retained and cleaned up later by Kubernetes.
## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/crds/embedded/virtualmachines.yaml
+++ b/crds/embedded/virtualmachines.yaml
@@ -3868,6 +3868,8 @@ spec:
                                   description: Path defines the path to disk file
                                     in the container
                                   type: string
+                                hotpluggable:
+                                  type: boolean
                               required:
                               - image
                               type: object
@@ -4491,6 +4493,35 @@ spec:
                           description: VolumeSource represents the source of the volume
                             to map to the disk.
                           properties:
+                            containerDisk:
+                              type: object
+                              description: |-
+                                ContainerDisk references a docker image, embedding a qcow or raw disk.
+                                More info: https://kubevirt.gitbooks.io/user-guide/registry-disk.html
+                              properties:
+                                image:
+                                  description: Image is the name of the image with
+                                    the embedded disk.
+                                  type: string
+                                imagePullPolicy:
+                                  description: |-
+                                    Image pull policy.
+                                    One of Always, Never, IfNotPresent.
+                                    Defaults to Always if :latest tag is specified, or IfNotPresent otherwise.
+                                    Cannot be updated.
+                                    More info: https://kubernetes.io/docs/concepts/containers/images#updating-images
+                                  type: string
+                                imagePullSecret:
+                                  description: ImagePullSecret is the name of the
+                                    Docker registry secret required to pull the image.
+                                    The secret must already exist.
+                                  type: string
+                                path:
+                                  description: Path defines the path to disk file
+                                    in the container
+                                  type: string
+                                hotpluggable:
+                                  type: boolean
                             dataVolume:
                               description: |-
                                 DataVolume represents the dynamic creation a PVC for this volume as well as
--- a/images/virt-artifact/patches/README.md
+++ b/images/virt-artifact/patches/README.md
@@ -140,3 +140,47 @@ d8-cni-cilium ensures that once the label is removed from the target pod, only t
 - Do not add cpu-model nodeSelector for "kvm64" model. This selector prevents starting VMs as node-labeler ignores to labeling nodes with "kvm64" model.
 
 - Overwrite calculated model on migration, put back "kvm64" for Discovery and Features vmclass types.
+
+---
+#### `032-hotplug-container-disk.patch`
+
+Add Hotplug container-disk volumes.
+How `container-disk` and HotPlug Work
+The `container-disk` is a program written in C used within KubeVirt to facilitate the mounting of container-based disk images into virtual machines. Its core function is to start up and create a UNIX socket within a specific directory. The program terminates when the socket is removed or upon receiving a `SIGTERM` signal.
+
+##### Key Workflow: `container-disk`
+
+##### Initialization
+- A sidecar container, running the `container-disk` image, is created alongside the `virt-launcher` pod.
+- An init-container in the `virt-launcher` pod copies the `container-disk` program to a shared `emptyDir` volume. This setup allows the sidecar to execute the program.
+
+##### Socket Creation
+
+- The `container-disk` program creates a socket in the `emptyDir` volume.
+- This shared volume allows the `virt-handler` to locate the socket on the host machine at:  
+  `/var/lib/kubelet/pods/.../volumes/kubernetes.io~empty-dir/`.
+
+##### Socket Detection and Mounting
+
+- Upon detecting the socket, `virt-handler` identifies it as a `container-disk` volume and retrieves its parent mount point.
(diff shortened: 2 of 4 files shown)


crds/embedded/virtualmachines.yaml
```diff
@@ -4491,6 +4493,35 @@ spec:
                           description: VolumeSource represents the source of the volume
                             to map to the disk.
                           properties:
+                            containerDisk:
+                              type: object
+                              description: |-
+                                ContainerDisk references a docker image, embedding a qcow or raw disk.
+                                More info: https://kubevirt.gitbooks.io/user-guide/registry-disk.html
```
diafour, line 4500: Non-working link

diafour: Oh, thats embedded. ok.

images/virt-artifact/patches/028-hotplug-container-disk.patch
```diff
@@ -0,0 +1,1827 @@
...
+@@ -3307,9 +3307,6 @@ spec:
+     - jsonPath: .status.phase
+       name: Phase
+       type: string
+-    deprecated: true
+-    deprecationWarning: kubevirt.io/v1alpha3 is now deprecated and will be removed
+-      in a future release.
+     name: v1alpha3
+     schema:
+       openAPIV3Schema:
```
diafour, line 135: This change is not needed, we use kv  as embedded CRD with only one version "v1".

images/virt-artifact/patches/028-hotplug-container-disk.patch
```diff
@@ -0,0 +1,1827 @@
...
++        - apiGroups:
++          - subresources.virtualization.deckhouse.io
++          resources:
++          - virtualmachines/addvolume
++          - virtualmachines/removevolume
++          verbs:
++          - update
+         - apiGroups:
+           - subresources.kubevirt.io
+           resources:
```
diafour, line 153: ClusterServiceVersion is not used. It is for the OpenShift.

images/virt-artifact/patches/028-hotplug-container-disk.patch
```diff
@@ -0,0 +1,1827 @@
...
++- apiGroups:
++  - subresources.virtualization.deckhouse.io
++  resources:
++  - virtualmachines/addvolume
++  - virtualmachines/removevolume
++  verbs:
++  - update
+ - apiGroups:
+   - subresources.kubevirt.io
+   resources:
```
diafour, line 180: Move these changes to `templates/kubevirt/virt-operator/rbac-for-us.yaml`

images/virt-artifact/patches/README.md
```diff
@@ -110,3 +110,7 @@ How does it work?
 2. When changing the affinity or nodeSelector in vmi, the vmi controller will set the `NodePlacementNotMatched` condition to True in vmi.
 3. The workload-updater controller monitors the vmi and starts migration when there is a `NodePlacementNotMatched` conditions on the vmi.
 4. When the migration is completed, virt-handler will remove the condition `NodePlacementNotMatched` from the vmi 
+
+#### `028-hotplug-container-disk.patch`
+
+Hotplug container-disk volumes.
```
diafour, line 116: Please, give more detail on how container disk works with hotplug.

Also, there is a new flag "userid", add details on this flag too.

diafour:
If there are some modifications over the upstream code, it'd be better to split into 2 patches: 1st with the untouched code from the upstream and 2nd with modifications from us.

images/virt-artifact/patches/028-hotplug-container-disk.patch
yaroslavborbat: It is generated automatically. I think, i will remove the generated directory from the patch

images/virt-artifact/patches/028-hotplug-container-disk.patch
yaroslavborbat: It is generated automatically. I think, i will remove the generated directory from the patch

images/virt-artifact/patches/README.md
yaroslavborbat: Yes. I have a plan to add more details. We currently have internal documentation with information about how this patch works.

images/virt-artifact/patches/029-hotplug-container-disk.patch
```diff
@@ -0,0 +1,1827 @@
...
+-	if len(attachmentPod.Spec.Volumes)-2 != len(volumes) {
++	// -3 if exist container-disk
++	magicNum := len(attachmentPod.Spec.Volumes) - 2
++	for _, volume := range volumes {
++		if volume.ContainerDisk != nil {
++			magicNum -= 1
++			break
++		}
++	}
++	if magicNum != len(volumes) {
```
diafour, line 465:
```suggestion
 	// -2 for empty dir and token
-	if len(attachmentPod.Spec.Volumes)-2 != len(volumes) {
+	// -3 if exist container-disk
+	podVolumesCount := len(attachmentPod.Spec.Volumes) - 2
+	for _, volume := range volumes {
+		if volume.ContainerDisk != nil {
+			podVolumesCount -= 1
+			break
+		}
+	}
+	if podVolumesCount != len(volumes) {
```

fl64 (approved):

Isteb4k (approved):

PR #569, opened 2024-12-11
feat(templates): add restricted access policy

yaroslavborbat:
## Description
add restricted access policy

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/templates/admission-policy.yaml
+++ b/templates/admission-policy.yaml
@@ -1,0 +1,67 @@
+{{- $kubeVersion := .Values.global.clusterConfiguration.kubernetesVersion }}
+{{- $apiVersion := "" }}
+{{- if semverCompare ">=1.30.0" $kubeVersion }}
+{{- $apiVersion = "admissionregistration.k8s.io/v1" }}
+{{- else if semverCompare ">=1.28.0" $kubeVersion }}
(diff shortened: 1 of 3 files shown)


Isteb4k (dismissed):

templates/admission-policy.yaml
```diff
@@ -0,0 +1,56 @@
+{{- $kubeVersion := semverCompare ">1.30.0" .Values.global.clusterConfiguration.kubernetesVersion }}
```
diafour, line 1: "kubeVersion" name has too wide meaning and no indication that it is actually a boolean flag.

templates/admission-policy.yaml
```diff
@@ -0,0 +1,56 @@
...
+apiVersion: admissionregistration.k8s.io/{{ if $kubeVersion }}v1{{ else }}v1alpha1{{ end }}
```
diafour, line 2: "admissionregistration.k8s.io/v1 for >1.30" seems strange. v1 appears in 1.30. 1.28 and 1.29 contains v1beta1 version for ValidatingAdmissionPolicy and ValidatingAdmissionPolicyBinding  (changelog recommends migrate from v1alpha to v1beta when upgrading from 1.27 to 1.28)

diafour (approved):

PR #577, opened 2024-12-16
refactor(vd): events

danilrwx:
## Description
Adding VirtualDisk events

Adding events is aimed at improving problem finding and tracking the causes of state changes

New events on VirtualDisk resources

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_cvi.go
+++ b/images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_cvi.go
@@ -105,12 +105,7 @@ func (ds ObjectRefClusterVirtualImage) Sync(ctx context.Context, vd *virtv2.Virt
 
 	switch {
 	case isDiskProvisioningFinished(condition):
-		ds.recorder.Event(
-			vd,
-			corev1.EventTypeNormal,
-			v1alpha2.ReasonDataSourceDiskProvisioningCompleted,
-			"Disk provisioning finished: clean up",
-		)
+		log.Debug("Disk provisioning finished: clean up")
 
 		setPhaseConditionForFinishedDisk(pvc, cb, &vd.Status.Phase, supgen)
 
--- a/images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_cvi.go
+++ b/images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_cvi.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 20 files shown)


images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_cvi.go
```diff
@@ -99,7 +105,12 @@ func (ds ObjectRefClusterVirtualImage) Sync(ctx context.Context, vd *virtv2.Virt
 
 	switch {
 	case isDiskProvisioningFinished(condition):
-		log.Debug("Disk provisioning finished: clean up")
+		ds.recorder.Event(
```
Isteb4k, line 108: No need to send event here

images/virtualization-artifact/pkg/controller/vd/internal/source/object_ref_vdsnapshot.go
```diff
@@ -75,7 +79,12 @@ func (ds ObjectRefVirtualDiskSnapshot) Sync(ctx context.Context, vd *virtv2.Virt
 
 	switch {
 	case isDiskProvisioningFinished(condition):
-		log.Debug("Disk provisioning finished: clean up")
+		ds.recorder.Event(
+			vd,
+			corev1.EventTypeNormal,
+			v1alpha2.ReasonDataSourceDiskProvisioningCompleted,
+			"Disk provisioning finished: clean up",
```
Isteb4k, line 86: No need to send event here.
Not only here, for other data sources as well

images/virtualization-artifact/pkg/controller/vd/internal/source/registry.go
```diff
@@ -161,7 +176,12 @@ func (ds RegistryDataSource) Sync(ctx context.Context, vd *virtv2.VirtualDisk) (
 
 		return reconcile.Result{Requeue: true}, nil
 	case !podutil.IsPodComplete(pod):
-		log.Info("Provisioning to DVCR is in progress", "podPhase", pod.Status.Phase)
```
Isteb4k, line 164: Please, keep logs like this one. They are wanted.

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vd/internal/source/registry.go
danilrwx: Sending events also writes to the log, surely it is necessary to duplicate logs?

Isteb4k: Ah, then it's ok :)

Isteb4k (approved):

Isteb4k (approved):

PR #586, opened 2024-12-18
feat(vd): add info represents quota exceed state

eofff:
## Description
Add info indicates that the quota of project is exceeded.

## Checklist
- [x] e2e tests passed.


images/virtualization-artifact/pkg/controller/vd/internal/source/upload.go
```diff
@@ -311,6 +319,14 @@ func (ds UploadDataSource) Sync(ctx context.Context, vd *virtv2.VirtualDisk) (re
...
 			return reconcile.Result{}, err
 		}
+
+		if pvc.Status.Phase == "Pending" {
+			cb.
+				Status(metav1.ConditionFalse).
+				Reason(vdcondition.Provisioning).
+				Message("PVC in pending state, if this continues for a long time, check the status of PVC and DataVolume manually.")
+		}
+
```
LopatinDmitr, line 329: How about combining this logic into a single function? 

We first set the conditioner inside the function and then modify it.

Isteb4k (approved):

PR #604, opened 2024-12-23
refactor(vmop): events

danilrwx:
## Description
Adding VMOP events

Adding events is aimed at improving problem finding and tracking the causes of state changes

New events on VMOP resources

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
+++ b/images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
@@ -74,7 +74,7 @@ func (h LifecycleHandler) Handle(ctx context.Context, s state.VMOperationState) 
 
 	// Initialize new VMOP resource: set label with vm name, set phase to Pending and all conditions to Unknown.
 	if changed.Status.Phase == "" {
-		h.recorder.Event(changed, corev1.EventTypeNormal, virtv2.ReasonVMOPStarted, "VMOP started")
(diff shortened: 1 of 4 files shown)


images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -70,6 +74,7 @@ func (h LifecycleHandler) Handle(ctx context.Context, s state.VMOperationState)
 
 	// Initialize new VMOP resource: set label with vm name, set phase to Pending and all conditions to Unknown.
 	if changed.Status.Phase == "" {
+		h.recorder.Event(changed, corev1.EventTypeNormal, virtv2.ReasonVMOPStarted, "VMOP started")
```
Isteb4k, line 77: It is better to avoid abbreviations in messages to the user.

images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -176,6 +189,7 @@ func (h LifecycleHandler) checkOperationComplete(ctx context.Context, changed *v
 
 	if isComplete {
 		changed.Status.Phase = virtv2.VMOPPhaseCompleted
+		h.recorder.Event(changed, corev1.EventTypeNormal, virtv2.ReasonVMOPSucceeded, "VMOP succeeded")
```
Isteb4k, line 192: It is better to avoid abbreviations in messages to the user.

Isteb4k (changes requested):

Isteb4k (approved):

PR #614, opened 2025-01-06
fix(module) fix register module issue

fl64:
## Description
Fix register module issue

We need to fix an issue:
```
register hooks search and register hooks: module hook --config invalid: get shell hook config: exit status 1
```

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/werf.yaml
+++ b/werf.yaml
@@ -120,12 +120,15 @@ shell:
       cat images_digests.json
 ---
 image: python-dependencies
-from: python:3.9-slim
+from: {{ .Images.BASE_ALT_P11 }}
 fromCacheVersion: "2024-11-07.1"
 git:
   - add: /lib/python/requirements.txt
     to: /requirements.txt
 shell:
+  beforeInstall:
+    - apt-get update
+    - apt-get install -y python3 python3-module-pip-run
   setup:
     - pip3 install -r /requirements.txt -t /dist
 ---


werf.yaml
```diff
@@ -120,12 +120,15 @@ shell:
...
-from: python:3.9-slim
+from: {{ .Images.BASE_ALT_P11 }}
 fromCacheVersion: "2024-11-07.1"
 git:
   - add: /lib/python/requirements.txt
     to: /requirements.txt
 shell:
+  beforeInstall:
+    - apt-get update
+    - apt-get install -y python3 python3-module-pip-run
```
universal-itengineer, line 131: ALT Linux, when installing the package this way we will get version 3.12.7 of python. Is this correct?

fl64: why not?

werf.yaml
```diff
@@ -120,12 +120,15 @@ shell:
       cat images_digests.json
 ---
 image: python-dependencies
-from: python:3.9-slim
+from: {{ .Images.BASE_ALT_P11 }}
 fromCacheVersion: "2024-11-07.1"
 git:
   - add: /lib/python/requirements.txt
     to: /requirements.txt
```
universal-itengineer, line 127:
```suggestion
git:
  - add: /lib/python/requirements.txt
    to: /requirements.txt
    stageDependencies:
      beforeInstall:
        - '**/*'
```
For this case, it's better to use `stageDependencies` rager than `fromCacheVersion`

werf.yaml
yaroslavborbat: I only remember that we needed a version greater than 3.8 ( using `:=` )

werf.yaml
fl64:
```
Error: unable to generate default repo: unable to load werf config: unknown fields: `beforeInstall`!

    beforeInstall:
    - '**/*'

/tmp/werf-config-render-3021092801
```

yaroslavborbat (approved):

PR #625, opened 2025-01-14
fix(vd): fix condition status updates in VirtualDisk

LopatinDmitr:
## Description

This PR addresses two issues related to the update generation in conditions for the VirtualDisk module. It fixes the update generation process and changes the status of in-use conditions from "Unknown" to "False".

By fixing the update generation and clarifying the status of in-use conditions, this PR ensures that the system behaves predictably and accurately reflects the actual state of conditions.

After these changes, condition statuses will be updated correctly, and in-use conditions will no longer show as "Unknown".
## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
+++ b/images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
@@ -151,39 +151,39 @@ func (h InUseHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (recon
 
 	cb := conditions.NewConditionBuilder(vdcondition.InUseType)
 	switch {
-	case allowUseForVM && inUseCondition.Status == metav1.ConditionUnknown:
-		if inUseCondition.Reason != vdcondition.AllowedForVirtualMachineUsage.String() {
+	case allowUseForVM && inUseCondition.Status != metav1.ConditionTrue:
+		if inUseCondition.Reason != vdcondition.AttachedToVirtualMachine.String() {
 			cb.
 				Generation(vd.Generation).
 				Status(metav1.ConditionTrue).
-				Reason(vdcondition.AllowedForVirtualMachineUsage).
+				Reason(vdcondition.AttachedToVirtualMachine).
 				Message("")
 			conditions.SetCondition(cb, &vd.Status.Conditions)
 			return reconcile.Result{}, nil
 		}
-	case allowUseForImage && inUseCondition.Status == metav1.ConditionUnknown:
-		if inUseCondition.Reason != vdcondition.AllowedForImageUsage.String() {
+	case allowUseForImage && inUseCondition.Status != metav1.ConditionTrue:
+		if inUseCondition.Reason != vdcondition.UsedForImageCreation.String() {
 			cb.
 				Generation(vd.Generation).
 				Status(metav1.ConditionTrue).
-				Reason(vdcondition.AllowedForImageUsage).
+				Reason(vdcondition.UsedForImageCreation).
 				Message("")
 			conditions.SetCondition(cb, &vd.Status.Conditions)
 			return reconcile.Result{}, nil
 		}
 	default:
-		setUnknown := false
+		setNotInUse := false
 
-		if inUseCondition.Reason == vdcondition.AllowedForVirtualMachineUsage.String() && !allowUseForVM {
-			setUnknown = true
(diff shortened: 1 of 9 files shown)


images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
```diff
@@ -143,42 +151,49 @@ func (h InUseHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (recon
...
-				Reason(vdcondition.AllowedForVirtualMachineUsage).
+				Reason(vdcondition.AttachedToVirtualMachine).
 				Message("")
 			conditions.SetCondition(cb, &vd.Status.Conditions)
+			return reconcile.Result{}, nil
 		}
-	case allowUseForImage && inUseCondition.Status == metav1.ConditionUnknown:
-		if inUseCondition.Reason != vdcondition.AllowedForImageUsage.String() {
+	case allowUseForImage && inUseCondition.Status != metav1.ConditionTrue:
+		if inUseCondition.Reason != vdcondition.UsedForImageCreation.String() {
```
Isteb4k, line 165: Suggest deleting this `if` statement, don't see the reason to keep it

images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
```diff
@@ -143,42 +151,49 @@ func (h InUseHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (recon
 
 	cb := conditions.NewConditionBuilder(vdcondition.InUseType)
 	switch {
-	case allowUseForVM && inUseCondition.Status == metav1.ConditionUnknown:
-		if inUseCondition.Reason != vdcondition.AllowedForVirtualMachineUsage.String() {
+	case allowUseForVM && inUseCondition.Status != metav1.ConditionTrue:
```
Isteb4k, line 154: In fact, `allowUseForVM` variable doesn’t indicate whether it allowed for a virtual machine; it shows whether the disk is being used by a virtual machine. Maybe rename it to usedByVM/usedByImage ?

images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
```diff
@@ -143,42 +151,49 @@ func (h InUseHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (recon
...
+		if inUseCondition.Reason != vdcondition.AttachedToVirtualMachine.String() {
```
Isteb4k, line 155: Suggest deleting this if statement, don't see the reason to keep it

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
LopatinDmitr: Renamed.

images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vd/internal/inuse.go
```diff
@@ -143,42 +151,49 @@ func (h InUseHandler) Handle(ctx context.Context, vd *virtv2.VirtualDisk) (recon
...
+		if inUseCondition.Reason == vdcondition.UsedForImageCreation.String() && !allowUseForImage {
+			setNotInUse = true
 		}
 
-		if setUnknown {
-			cb.Generation(vd.Generation).Status(metav1.ConditionUnknown).Reason(conditions.ReasonUnknown).Message("")
+		if setNotInUse {
+			cb.Generation(vd.Generation).Status(metav1.ConditionFalse).Reason(vdcondition.NotInUse).Message("")
 			conditions.SetCondition(cb, &vd.Status.Conditions)
 			return reconcile.Result{Requeue: true}, nil
```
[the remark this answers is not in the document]
LopatinDmitr: Currently, the requeue in this place is used to implement the logic of first resetting the condition to False from the previous state and only then setting it to True (if needed).

We need to redesign the handler's operation to change the logic for updating the Reason on the fly without intermediate switching to False and remove the requeue.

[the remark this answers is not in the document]
LopatinDmitr: Added TODO.

Isteb4k (approved):

PR #678, opened 2025-01-29
fix(vm): fix VM restart behavior with invalid spec

LopatinDmitr:
## Description

This change introduces a "Pending" phase for virtual machines (VMs) that have an invalid specification and are in the restart process. Previously, if a VM was rebooted while having an invalid spec, it would start with the old spec.

This update is necessary to ensure that VMs do not continue to operate with outdated or invalid configurations after a restart.

With this change, when a VM with an invalid spec is restarted, it will enter a "Pending" phase, allowing for corrective actions before proceeding with the boot process. 

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: vm
type: fix
summary: move the vm to a pending phase instead of starting it if it has invalid specs during start/restart to prevent the use of outdated specifications
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
@@ -22,16 +22,20 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	virtv1 "kubevirt.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/deckhouse/virtualization-controller/pkg/common/annotations"
 	kvvmutil "github.com/deckhouse/virtualization-controller/pkg/common/kvvm"
+	"github.com/deckhouse/virtualization-controller/pkg/controller/conditions"
 	"github.com/deckhouse/virtualization-controller/pkg/controller/powerstate"
 	"github.com/deckhouse/virtualization-controller/pkg/controller/vm/internal/state"
 	"github.com/deckhouse/virtualization-controller/pkg/eventrecord"
 	"github.com/deckhouse/virtualization-controller/pkg/logger"
 	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
+	"github.com/deckhouse/virtualization/api/core/v1alpha2/vmcondition"
 )
 
 const nameSyncPowerStateHandler = "SyncPowerStateHandler"
@@ -104,6 +108,9 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 		shutdownInfo = s.ShutdownInfo
 	})
 
+	appliedCondtion, _ := conditions.GetCondition(vmcondition.TypeConfigurationApplied, s.VirtualMachine().Changed().Status.Conditions)
+	canStartVM := appliedCondtion.Status == metav1.ConditionTrue
+
 	switch runPolicy {
 	case virtv2.AlwaysOffPolicy:
 		if kvvmi != nil {
@@ -120,6 +127,10 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 		}
 	case virtv2.AlwaysOnPolicy:
 		if kvvmi == nil {
+			if !canStartVM {
+				return h.handleCanNotStartVM(ctx, s, kvvm)
+			}
+
 			h.recordStartEventf(ctx, s.VirtualMachine().Current(),
 				"Start initiated by controller for %v policy",
 				runPolicy,
@@ -132,6 +143,10 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 
 		if kvvmi != nil && kvvmi.DeletionTimestamp == nil {
 			if kvvmi.Status.Phase == virtv1.Succeeded {
+				if !canStartVM {
+					return h.handleCanNotStartVM(ctx, s, kvvm)
+				}
+
 				if shutdownInfo.PodCompleted {
 					// Treat completed Pod as restart if guest was restarted or as a start if guest was stopped.
 					switch shutdownInfo.Reason {
@@ -157,6 +172,10 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 			}
 
 			if kvvmi.Status.Phase == virtv1.Failed {
+				if !canStartVM {
+					return h.handleCanNotStartVM(ctx, s, kvvm)
+				}
+
 				h.recordRestartEventf(ctx, s.VirtualMachine().Current(),
 					"Restart initiated by controller for %s runPolicy after observing failed guest VM",
 					runPolicy,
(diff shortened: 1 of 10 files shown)


images/virtualization-artifact/pkg/controller/vm/internal/util.go
```diff
@@ -66,69 +70,120 @@ func conditionStatus(status string) metav1.ConditionStatus {
...
-	return mapPhases[kvvm.Status.PrintableStatus]
+	if handler, exists := mapPhases[kvvm.Status.PrintableStatus]; exists {
+		return handler(vm, kvvm)
+	}
+
+	return virtv2.MachinePending
 }
 
-var mapPhases = map[virtv1.VirtualMachinePrintableStatus]virtv2.MachinePhase{
+type PhaseHandler func(...interface{}) virtv2.MachinePhase
```
yaroslavborbat, line 96: The use of PhaseHandler as func(...interface{}) virtv2.MachinePhase doesn't provide any real benefit in this context. The idea behind variadic interface{} parameters is typically to allow flexibility in argument passing, but here, every handler is always called through a mapping (mapPhases), meaning the caller doesn’t know which specific function will be executed.

This makes the flexible signature pointless because:

The arguments are always the same (vm and kvvm). There’s no scenario where a handler would need a different set of parameters.
Since the caller doesn't know which function will be executed, it must always pass the same arguments, making variadic interface{} unnecessary.
Each handler still assumes a fixed set of arguments and immediately type-asserts them, which means the flexibility isn't even being used.

yaroslavborbat: Additionally, I recommend renaming PhaseHandler to PhaseGetter. The current name suggests that these functions handle something, but in reality, they simply return (get) a phase based on the provided VM state.

images/virtualization-artifact/pkg/controller/vm/internal/util.go
```diff
@@ -66,69 +70,120 @@ func conditionStatus(status string) metav1.ConditionStatus {
...
+		}
+
+		return virtv2.MachineStopped
+	},
 	// VirtualMachineStatusProvisioning indicates that cluster resources associated with the virtual machine
 	// (e.g., DataVolumes) are being provisioned and prepared.
-	virtv1.VirtualMachineStatusProvisioning: virtv2.MachineStarting,
+	virtv1.VirtualMachineStatusProvisioning: func(args ...interface{}) virtv2.MachinePhase {
+		return virtv2.MachineStarting
+	},
```
yaroslavborbat, line 118: Since this handler does not use args, it would be clearer to replace it with _ to explicitly indicate that the arguments are ignored

yaroslavborbat (changes requested):

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -68,9 +72,26 @@ func (h *SyncPowerStateHandler) Handle(ctx context.Context, s state.VirtualMachi
 		log.Error(err.Error())
 	}
 
+	err = h.syncKVVMAnnotations(ctx, changed, kvvm)
```
Isteb4k, line 75: Lost the error above: it will never be returned, though it should be.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -104,6 +125,9 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 		shutdownInfo = s.ShutdownInfo
 	})
 
+	appliedCondtion, _ := conditions.GetCondition(vmcondition.TypeConfigurationApplied, s.VirtualMachine().Changed().Status.Conditions)
```
Isteb4k, line 128: typo

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -166,10 +202,23 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
...
+			h.recordStartEventf(ctx, s.VirtualMachine().Current(),
+				"Start initiated by controller for %v policy",
+				runPolicy,
+			)
+
+			if err = powerstate.StartVM(ctx, h.client, kvvm); err != nil {
+				return fmt.Errorf("failed to start VM: %w", err)
+			}
 		}
 	case virtv2.AlwaysOnUnlessStoppedManually:
```
Isteb4k, line 215: Terrible cyclomatic complexity. Very hard to comprehend; it needs refactoring and unit test coverage.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -104,6 +125,9 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 		shutdownInfo = s.ShutdownInfo
 	})
 
+	appliedCondtion, _ := conditions.GetCondition(vmcondition.TypeConfigurationApplied, s.VirtualMachine().Changed().Status.Conditions)
+	canStartVM := appliedCondtion.Status == metav1.ConditionTrue
```
Isteb4k, line 129: Rename canStartVM to isConfigurationApplied, or something like this

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -246,12 +313,32 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
...
+			if err = powerstate.StartVM(ctx, h.client, kvvm); err != nil {
+				return fmt.Errorf("failed to start VM: %w", err)
+			}
 		}
 	}
 
 	return nil
 }
 
+func (h *SyncPowerStateHandler) handleCanNotStartVM(ctx context.Context, s state.VirtualMachineState, kvvm *virtv1.VirtualMachine) error {
```
Isteb4k, line 331: A method should have a name that reflects what it does internally, rather than the scenario/branch in which it is used.

images/virtualization-artifact/pkg/controller/vm/internal/util.go
```diff
@@ -192,3 +247,42 @@ func isInternalVirtualMachineError(phase virtv1.VirtualMachinePrintableStatus) b
 func podFinal(pod corev1.Pod) bool {
 	return pod.Status.Phase == corev1.PodSucceeded || pod.Status.Phase == corev1.PodFailed
 }
+
+func addStartAnnotationToKVVM(ctx context.Context, cl client.Client, kvvm *virtv1.VirtualMachine) error {
```
Isteb4k, line 251: Move all the code that interacts with this annotation into a single file. Don't use util as a catch-all for everything.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
```diff
@@ -132,6 +160,10 @@ func (h *SyncPowerStateHandler) syncPowerState(ctx context.Context, s state.Virt
 
 		if kvvmi != nil && kvvmi.DeletionTimestamp == nil {
 			if kvvmi.Status.Phase == virtv1.Succeeded {
```
Isteb4k, line 162: It's not clear from the code that Succeeded means we intend to start kvvm. It actually seems incorrect. What if vm was stopped internally?

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vm/internal/util.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/util.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/util.go
LopatinDmitr: Fixed in another PR inherited from this one.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
LopatinDmitr: Fixed in another PR inherited from this one.

images/virtualization-artifact/pkg/controller/vm/internal/sync_power_state.go
LopatinDmitr: Fixed.

Isteb4k (dismissed):

Isteb4k (approved):

yaroslavborbat (approved):

PR #690, opened 2025-02-03
fix(kubevirt): migration disk serial

yaroslavborbat:
## Description
Previously, we set the disk serial to match the disk name, relying on Kubernetes to ensure the uniqueness of names. However, after this [commit](https://github.com/qemu/qemu/commit/75997e182b695f2e3f0a2d649734952af5caf3ee) on QEMU, we can no longer do this, as QEMU no longer truncates the serial to 36 characters and now returns an error. We must handle the truncation ourselves. However, truncation does not guarantee uniqueness, so we switched to generating an MD5 hash of the disk name. The MD5 hash is easy to reproduce and fits the required 32-character length.

To ensure a successful migration, existing `kvvm` and `kvvmi` resources need to be patched to reflect the new serial format.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: vm  
type: fix  
summary: Disk serials are now generated using the MD5 hash of the disk name instead of the disk name itself. This prevents errors caused by recent QEMU changes enforcing a strict 36-character limit on serial numbers.
```

### DIFF
--- a/images/virtualization-artifact/pkg/migration/migration.go
+++ b/images/virtualization-artifact/pkg/migration/migration.go
@@ -24,7 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/deckhouse/deckhouse/pkg/log"
-
 	"github.com/deckhouse/virtualization-controller/pkg/logger"
 )
 
--- a/images/virtualization-artifact/pkg/migration/migration.go
+++ b/images/virtualization-artifact/pkg/migration/migration.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2025 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package migration
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/deckhouse/deckhouse/pkg/log"
+
+	"github.com/deckhouse/virtualization-controller/pkg/logger"
+)
+
+type constructor func(client client.Client, logger *log.Logger) (Migration, error)
+
+var newMigrations = []constructor{
+	newQEMUMaxLength36,
+}
+
+type Migration interface {
+	Name() string
+	Migrate(ctx context.Context) error
+}
+
+type Controller struct {
+	client client.Client
+	log    *log.Logger
+
+	migrations []Migration
+}
+
+func NewController(client client.Client, log *log.Logger) (*Controller, error) {
+	migrations := make([]Migration, len(newMigrations))
+	for i, newMigration := range newMigrations {
+		m, err := newMigration(client, log)
+		if err != nil {
+			return nil, err
+		}
+		migrations[i] = m
+	}
+
+	return &Controller{
+		client:     client,
+		log:        log,
+		migrations: migrations,
+	}, nil
+}
+
+func (c *Controller) Run(ctx context.Context) {
+	wg := &sync.WaitGroup{}
+
+	c.run(ctx, wg, c.migrations)
+	wg.Wait()
+}
+
+func (c *Controller) run(ctx context.Context, wg *sync.WaitGroup, migrations []Migration) {
+	for _, m := range migrations {
+		if wg != nil {
+			wg.Add(1)
+		}
+		lg := c.log.With("name", m.Name())
+		lg.Info("Running migration")
+
+		go func() {
+			if wg != nil {
+				defer wg.Done()
+			}
+
+			for {
+				select {
+				case <-ctx.Done():
+					lg.Info("Cancelled migration")
+					return
+				default:
+					if err := m.Migrate(ctx); err != nil {
+						lg.Error("Failed to run migration, retry after 5s...", logger.SlogErr(err))
+						time.Sleep(5 * time.Second)
+						continue
+					}
+					lg.Info("Finished migration")
+					return
+				}
+			}
+		}()
+	}
+}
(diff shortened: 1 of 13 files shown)


images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
```diff
@@ -0,0 +1,116 @@
+/*
```
Isteb4k, line 1: Wrong file naming

images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/deckhouse/deckhouse/pkg/log"
+	virtv1 "kubevirt.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/deckhouse/virtualization-controller/pkg/controller/kvbuilder"
+	"github.com/deckhouse/virtualization-controller/pkg/logger"
+)
+
+const (
+	// https://github.com/qemu/qemu/commit/75997e182b695f2e3f0a2d649734952af5caf3ee
```
Isteb4k, line 31: Let's add a README where we describe in detail why we need each migration, similar to the README for patches.

images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for i := range kvvmList.Items {
+		kvvm := &kvvmList.Items[i]
+		if !r.needMigrate(&kvvm.Spec.Template.Spec) {
+			continue
+		}
+		copied := kvvm.DeepCopy()
+		r.mutateSpec(&copied.Spec.Template.Spec)
+		r.logger.Debug("Patch kvvm with new kvvmi spec", logger.SlogTryJson("newSpce", copied.Spec.Template.Spec))
```
Isteb4k, line 70: Let's set Info

images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for i := range kvvmiList.Items {
+		kvvmi := &kvvmiList.Items[i]
+		if !r.needMigrate(&kvvmi.Spec) {
+			continue
+		}
+		copied := kvvmi.DeepCopy()
+		r.mutateSpec(&copied.Spec)
+		r.logger.Debug("Patch kvvmi with new spec", logger.SlogTryJson("newSpce", copied.Spec))
```
Isteb4k, line 89: Info as well

images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
```diff
@@ -0,0 +1,116 @@
...
+		if err = r.client.Patch(ctx, kvvmi, client.MergeFrom(copied)); err != nil {
+			return err
+		}
+	}
+
```
Isteb4k, line 94: I’d like to see how many resources were affected.

images/virtualization-artifact/pkg/migration/migration.go
```diff
@@ -0,0 +1,109 @@
...
+			c.blockingMigrations = append(c.blockingMigrations, m)
+		} else {
+			c.nonBlockingMigrations = append(c.nonBlockingMigrations, m)
+		}
+	}
+	return c, nil
+}
+
+func (c *Controller) Run(ctx context.Context) {
+	run := func(wg *sync.WaitGroup, migrations []Migration) {
```
Isteb4k, line 73: Why is the function in scope? The advantages of this approach aren’t clear, unlike the issues.

images/virtualization-artifact/pkg/migration/migration.go
```diff
@@ -0,0 +1,109 @@
...
+func NewController(client client.Client, log *log.Logger) *Controller {
+	return &Controller{
+		client: client,
+		log:    log,
+	}
+}
+
+func (c *Controller) Complete() (*Controller, error) {
+	for _, fn := range newMigrations {
+		m, err := fn(c.client, c.log)
```
Isteb4k, line 59: Choose another name
It's completely unclear what’s hidden inside fn

images/virtualization-artifact/pkg/migration/migration.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func NewController(client client.Client, log *log.Logger) *Controller {
+	return &Controller{
+		client: client,
+		log:    log,
+	}
+}
+
+func (c *Controller) Complete() (*Controller, error) {
```
Isteb4k, line 57: Looks like we dont need this method 
Why not move this logic to NewController ?

images/virtualization-artifact/pkg/migration/migration.go
```diff
@@ -0,0 +1,109 @@
...
+	run := func(wg *sync.WaitGroup, migrations []Migration) {
+		for _, m := range migrations {
+			if wg != nil {
+				wg.Add(1)
+			}
+			lg := c.log.With("name", m.Name())
+			lg.Info("Running migration")
+
+			go func() {
+				defer lg.Info("Finished migration")
```
Isteb4k, line 82: Migration is not finished if ctx.Done.

images/virtualization-artifact/pkg/migration/migration.go
```diff
@@ -0,0 +1,109 @@
...
+	IsBlocking() bool
+	Migrate(ctx context.Context) error
+}
+
+type Controller struct {
+	client client.Client
+	log    *log.Logger
+
+	blockingMigrations    []Migration
+	nonBlockingMigrations []Migration
```
Isteb4k, line 47: Do we really need nonBlockingMigrations migrations now?

Isteb4k (changes requested):

images/virtualization-artifact/pkg/migration/migration.go
yaroslavborbat: not sure

images/virtualization-artifact/pkg/migration/qemu-max-length-36.go
yaroslavborbat: What do you mean? Do you want to see the entire list of patches with errors?

Isteb4k (approved):

Isteb4k (approved):

PR #709, opened 2025-02-06
feat(e2e): verifying vm launching by qemu agent ready status

eofff:
## Description
This pull request enhances the reliability of our end-to-end (E2E) tests by updating the verification process for virtual machine (VM) readiness. Instead of merely waiting for the VM to reach the "running" phase, the test now explicitly checks the VM's status and optionally confirms the successful launch of the QEMU agent. This change aims to reduce false positives and ensure that the VM is fully operational and ready for subsequent testing steps. Adjustments are reflected in the affected test scripts to incorporate these additional validation checks.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: e2e
type: feat
summary:  verifying vm launching by qemu agent ready status
```

### DIFF
--- a/tests/e2e/affinity_toleration_test.go
+++ b/tests/e2e/affinity_toleration_test.go
@@ -98,8 +98,8 @@ var _ = Describe("Virtual machine affinity and toleration", ginkgoutil.CommonE2E
 
 	Context("When virtual machines are applied:", func() {
 		It("checks VMs phases", func() {
-			By(fmt.Sprintf("VMs should be in %s phases", PhaseRunning))
-			WaitPhaseByLabel(kc.ResourceVM, PhaseRunning, kc.WaitOptions{
+			By("VMs should be running")
+			WaitVmRunning(true, kc.WaitOptions{
 				Labels:    testCaseLabel,
 				Namespace: conf.Namespace,
 				Timeout:   MaxWaitTimeout,
--- a/tests/e2e/affinity_toleration_test.go
+++ b/tests/e2e/affinity_toleration_test.go
@@ -98,7 +98,7 @@ var _ = Describe("Virtual machine affinity and toleration", ginkgoutil.CommonE2E
 
 	Context("When virtual machines are applied:", func() {
 		It("checks VMs phases", func() {
-			By("VMs should be running")
+			By("VMs should be ready")
 			WaitVmReady(kc.WaitOptions{
 				Labels:    testCaseLabel,
 				Namespace: conf.Namespace,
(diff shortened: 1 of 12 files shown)


tests/e2e/util_test.go
```diff
@@ -253,6 +253,20 @@ func ChmodFile(pathFile string, permission os.FileMode) {
 	}
 }
 
+func WaitVmRunning(waitAgent bool, opts kc.WaitOptions) {
```
hardcoretime, line 256: WaitVmReady() - this seems better because `Running` is already used in the `phase` field.  
WaitVmReady(opts kc.WaitOptions) - I think `waitAgent` shouldn't be an argument because `Agent` should always be in the `ready` status in this function.

tests/e2e/vd_snapshots_test.go
```diff
@@ -316,8 +316,8 @@ var _ = Describe("Virtual disk snapshots", ginkgoutil.CommonE2ETestDecorators(),
 
 	Context("When virtual machines are applied:", func() {
 		It("checks VMs phases", func() {
-			By(fmt.Sprintf("VMs should be in %s phases", PhaseRunning))
-			WaitPhaseByLabel(kc.ResourceVM, PhaseRunning, kc.WaitOptions{
+			By("VMs should be running")
```
hardcoretime, line 319: running -> ready

tests/e2e/tests_suite_test.go
```diff
@@ -49,6 +49,7 @@ const (
 	PhaseRunning              = "Running"
 	PhaseWaitForUserUpload    = "WaitForUserUpload"
 	PhaseWaitForFirstConsumer = "WaitForFirstConsumer"
+	AgentReadyCondition       = "AgentReady"
```
hardcoretime, line 52: We should use conditions from the API in tests:
https://github.com/deckhouse/virtualization/blob/main/api/core/v1alpha2/vmcondition/condition.go#L34

tests/e2e/affinity_toleration_test.go
```diff
@@ -98,8 +98,8 @@ var _ = Describe("Virtual machine affinity and toleration", ginkgoutil.CommonE2E
 
 	Context("When virtual machines are applied:", func() {
 		It("checks VMs phases", func() {
-			By(fmt.Sprintf("VMs should be in %s phases", PhaseRunning))
-			WaitPhaseByLabel(kc.ResourceVM, PhaseRunning, kc.WaitOptions{
+			By("VMs should be running")
+			WaitVmRunning(true, kc.WaitOptions{
```
hardcoretime, line 102: It is not necessary to use `WaitVmReady` in each test case because it increases the duration of the tests. We should use it only in test cases where we need to wait for the `AgentReady` status for a specific case.

hardcoretime (changes requested):
See comments.

hardcoretime (approved):

PR #743, opened 2025-02-13
feat(kubevirt): add patch get-applied-checksum

yaroslavborbat:
## Description
This patch introduces the GetAppliedChecksum() method in virt-launcher.

virt-handler now tracks synchronized VMIs and computes their checksums. Periodically, it queries virt-launcher for the applied checksum. This ensures that we have two checksums:

The last one sent for synchronization.
The one actually applied.
Storing these checksums in VMI annotations helps verify that spec changes were pushed by virt-handler and not by an attacker.

## Checklist
- [X] e2e tests passed.

## Changelog entries

```changes
section:  feat
type: core
summary: virt-handler now tracks synchronized VMIs and computes their checksums. Periodically, it queries virt-launcher for the applied checksum. This ensures that we have two checksums:

 
```


images/virt-artifact/patches/039-get-applied-checksum.patch
```diff
@@ -0,0 +1,1213 @@
...
+-	fmt "fmt"
+-
+-	proto "github.com/golang/protobuf/proto"
+-
+-	math "math"
++import proto "github.com/golang/protobuf/proto"
++import fmt "fmt"
++import math "math"
+ 
++import (
```
diafour, line 19: Why? =)

diafour (approved):

images/virt-artifact/patches/039-get-applied-checksum.patch
diafour: Generated code, may be a new version of the generator.

PR #750, opened 2025-02-14
test(images): add e2e image attachment test case

hardcoretime:
## Description

- refactor e2e disk attachment test case
- refactor e2e disk resizing test case

It fixes test cases where we use blockDeviceId and adds a new test case for the image attachment flow.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: images
type: test
summary: "It fixes test cases where we use blockDeviceId and adds a new test case for the image attachment flow."
```

### DIFF
--- a/tests/e2e/image_hotplug_test.go
+++ b/tests/e2e/image_hotplug_test.go
@@ -25,6 +25,7 @@ import (
 	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
 	"github.com/deckhouse/virtualization/tests/e2e/config"
 	d8 "github.com/deckhouse/virtualization/tests/e2e/d8"
+	"github.com/deckhouse/virtualization/tests/e2e/ginkgoutil"
 	kc "github.com/deckhouse/virtualization/tests/e2e/kubectl"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -92,7 +93,39 @@ func IsBlockDeviceReadOnly(vmName, blockDeviceId string) (bool, error) {
 	return roOpt == "ro", nil
 }
 
-var _ = Describe("Image hotplug", func() {
+type Counter struct {
+	Current  int
+	Expected int
+}
+
+type ReusableResources map[kc.Resource]*Counter
+
+// Useful when require to check the created resources in `REUSABLE` mode.
+//
+//	Static output option: `jsonpath='{.items[*].metadata.name}'`.
+func CheckReusableResources(resources ReusableResources, opts kc.GetOptions) {
+	GinkgoHelper()
+	opts.Output = "jsonpath='{.items[*].metadata.name}'"
+	for r, c := range resources {
+		res := kubectl.List(r, opts)
+		Expect(res.Error()).NotTo(HaveOccurred(), "failed to check the reusable resourse %q: %s", r, res.StdErr())
+		c.Current = len(strings.Split(res.StdOut(), " "))
+	}
+
+	isResourceCountExpected := false
+	for _, c := range resources {
+		if c.Current == c.Expected {
+			isResourceCountExpected = true
+		} else {
+			isResourceCountExpected = false
+		}
+	}
+	if isResourceCountExpected {
+		return
+	}
+}
+
+var _ = Describe("Image hotplug", ginkgoutil.CommonE2ETestDecorators(), func() {
 	const (
 		viCount    = 2
 		cviCount   = 2
@@ -103,80 +136,43 @@ var _ = Describe("Image hotplug", func() {
 	)
 
 	var (
-		vmObj             virtv2.VirtualMachine
-		disksBefore       Disks
-		disksAfter        Disks
-		testCaseLabel     = map[string]string{"testcase": "image-hotplug"}
-		isoLabel          = "iso"
-		imageBlockDevices = make([]Image, 0)
+		vmObj         virtv2.VirtualMachine
+		disksBefore   Disks
+		disksAfter    Disks
+		testCaseLabel = map[string]string{"testcase": "image-hotplug"}
 	)
 
-	Context("Preparing the environment", func() {
-		It("sets the namespace", func() {
-			kustomization := fmt.Sprintf("%s/%s", conf.TestData.ImageHotplug, "kustomization.yaml")
-			ns, err := kustomize.GetNamespace(kustomization)
-			Expect(err).NotTo(HaveOccurred(), "%w", err)
-			conf.SetNamespace(ns)
-		})
+	BeforeAll(func() {
+		kustomization := fmt.Sprintf("%s/%s", conf.TestData.ImageHotplug, "kustomization.yaml")
+		ns, err := kustomize.GetNamespace(kustomization)
+		Expect(err).NotTo(HaveOccurred(), "%w", err)
+		conf.SetNamespace(ns)
 	})
 
-	Context("When virtualization resources are applied", func() {
+	Context("When the virtualization resources are applied", func() {
 		It("result should be succeeded", func() {
 			if config.IsReusable() {
-				vms := &virtv2.VirtualMachineList{}
-				err := GetObjects(virtv2.VirtualMachineResource, vms, kc.GetOptions{
+				CheckReusableResources(ReusableResources{
+					virtv2.VirtualMachineResource: &Counter{
+						Expected: vmCount,
+					},
+					virtv2.VirtualDiskResource: &Counter{
+						Expected: vdCount,
+					},
+					virtv2.VirtualImageResource: &Counter{
+						Expected: viCount,
+					},
+					virtv2.ClusterVirtualImageResource: &Counter{
+						Expected: cviCount,
+					},
+					virtv2.VirtualMachineBlockDeviceAttachmentResource: &Counter{
+						Expected: vmbdaCount,
+					},
+				}, kc.GetOptions{
 					Labels:         testCaseLabel,
 					Namespace:      conf.Namespace,
 					IgnoreNotFound: true,
 				})
-				Expect(err).NotTo(HaveOccurred(), "failed to check reusable `VirtualMachine`", err)
-
-				vds := &virtv2.VirtualDiskList{}
-				err = GetObjects(virtv2.VirtualDiskResource, vds, kc.GetOptions{
-					Labels:         testCaseLabel,
-					Namespace:      conf.Namespace,
-					IgnoreNotFound: true,
-				})
-				Expect(err).NotTo(HaveOccurred(), "failed to check reusable `VirtualDisk`", err)
-
-				vis := &virtv2.VirtualImageList{}
-				err = GetObjects(virtv2.VirtualImageResource, vis, kc.GetOptions{
-					Labels:         testCaseLabel,
-					Namespace:      conf.Namespace,
-					IgnoreNotFound: true,
-				})
-				Expect(err).NotTo(HaveOccurred(), "failed to check reusable `VirtualImages`", err)
-
-				cvis := &virtv2.ClusterVirtualImageList{}
-				err = GetObjects(virtv2.ClusterVirtualImageResource, cvis, kc.GetOptions{
-					Labels:         testCaseLabel,
-					IgnoreNotFound: true,
-				})
-				Expect(err).NotTo(HaveOccurred(), "failed to check reusable `ClusterVirtualImages`", err)
-
-				vmbdas := &virtv2.VirtualMachineBlockDeviceAttachmentList{}
-				err = GetObjects(virtv2.VirtualMachineBlockDeviceAttachmentResource, vmbdas, kc.GetOptions{
-					Labels:         testCaseLabel,
-					IgnoreNotFound: true,
-				})
-				Expect(err).NotTo(HaveOccurred(), "failed to check reusable `VirtualMachineBlockDeviceAttachments`", err)
-
-				if len(vms.Items) == vmCount &&
-					len(vds.Items) == vdCount &&
-					len(vis.Items) == viCount &&
-					len(cvis.Items) == cviCount &&
-					len(vmbdas.Items) == vmbdaCount {
-					fmt.Println("all reusable resources have been found")
-					return
-				}
-
-				if len(vms.Items) == 0 &&
-					len(vds.Items) == 0 &&
-					len(vis.Items) == 0 &&
-					len(cvis.Items) == 0 &&
-					len(vmbdas.Items) == 0 {
-					fmt.Println("no found reusable resources in `REUSABLE` mode")
-				}
 			}
 
 			res := kubectl.Apply(kc.ApplyOptions{
@@ -185,10 +181,8 @@ var _ = Describe("Image hotplug", func() {
 			})
 			Expect(res.Error()).NotTo(HaveOccurred(), res.StdErr())
 		})
-	})
 
-	Context("When virtual images are applied", func() {
-		It("checks the `VirtualImages` phase", func() {
+		It("checks the resources phase", func() {
 			By(fmt.Sprintf("`VirtualImages` should be in the %q phase", virtv2.ImageReady), func() {
 				WaitPhaseByLabel(kc.ResourceVI, PhaseReady, kc.WaitOptions{
 					Labels:    testCaseLabel,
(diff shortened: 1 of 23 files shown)


tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+				if err != nil {
+					return unacceptableCount, err
+				}
+				diskCountAfter := len(disksAfter.BlockDevices)
+				return diskCountAfter, nil
+			}).WithTimeout(Timeout).WithPolling(Interval).Should(Equal(diskCountBefore), "comparing error: 'after' must be equal 'before'")
+		})
+	})
+
+	Context("When test is completed", func() {
```
yaroslavborbat, line 424: Use AfterAll

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	virtv1 "kubevirt.io/api/core/v1"
+)
+
+const (
+	viCount     = 2
+	cviCount    = 2
+	vmCount     = 1
+	vdCount     = 1
+	vmbdaCount  = 0
+	imgCount    = viCount + cviCount
```
yaroslavborbat, line 40: move to test body

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	res := d8Virtualization.SshCommand(vmName, cmd, d8.SshOptions{
+		Namespace:   conf.Namespace,
+		Username:    conf.TestData.SshUser,
+		IdenityFile: conf.TestData.Sshkey,
+	})
+	if res.Error() != nil {
+		return false, errors.New(res.StdErr())
+	}
+	options := strings.Split(res.StdOut(), ",")
+	roOpt := options[0]
```
yaroslavborbat, line 102: panic if slice is empty

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	return roOpt == readOnlyOpt, nil
```
yaroslavborbat, line 103: i think. You can hardcode "ro" and const is no needed

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	readOnlyOpt = "ro"
+)
+
+type Image struct {
+	Kind string
+	Name string
+}
+
+func IsBlockDeviceCdRom(vmName, blockDeviceName string) (bool, error) {
+	GinkgoHelper()
```
yaroslavborbat, line 51: GinkgoHelper() is unnecessary since this function has no Expect or other test assertions. It should be used in test-specific helpers for better error traceability.

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+	if len(blockDevices.BlockDevices) != 1 {
+		return false, fmt.Errorf("`blockDevices` length should be 1")
+	}
+	blockDevice := &blockDevices.BlockDevices[0]
+	return blockDevice.Type == cdRom, nil
+}
+
+func MountBlockDevice(vmName, blockDeviceId string) error {
+	GinkgoHelper()
```
yaroslavborbat, line 75: unnecessary

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+		IdenityFile: conf.TestData.Sshkey,
+	})
+	if res.Error() != nil {
+		return errors.New(res.StdErr())
+	}
+	return nil
+}
+
+func IsBlockDeviceReadOnly(vmName, blockDeviceId string) (bool, error) {
+	GinkgoHelper()
```
yaroslavborbat, line 90: unnecessary

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+	var (
+		vmObj         virtv2.VirtualMachine
+		disksBefore   Disks
+		disksAfter    Disks
+		testCaseLabel = map[string]string{"testcase": "image-hotplug"}
+		isoLabel      = "iso"
+		blockDevices  = make([]Image, 0)
+	)
+
+	Context("Preparing the environment", func() {
```
yaroslavborbat, line 116: use BeforeAll

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+				if _, ok := imgs[disk.Name]; ok {
+					if strings.HasSuffix(disk.Name, isoLabel) {
+						imgs[disk.Name] = fmt.Sprintf("%s-%s", CdRomIdPrefix, disk.Name)
+					} else {
+						imgs[disk.Name] = fmt.Sprintf("%s_%s", DiskIdPrefix, disk.Serial)
+					}
+				}
+			}
+
+			Expect(len(imgs)).To(BeNumerically("==", imgCount), "there are only %d `blockDevices` in this case", imgCount)
```
yaroslavborbat, line 389: Expect(imgs).To(HaveLen(imgCount))
or
Expect(len(imgs)).To(Equal(imgCount))

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+			imgs := make(map[string]string, 0)
+
+			cviImgs := &virtv2.ClusterVirtualImageList{}
+			err := GetObjects(virtv2.ClusterVirtualImageResource, cviImgs, kc.GetOptions{
+				Labels: testCaseLabel,
+			})
+			Expect(err).NotTo(HaveOccurred(), "failed to get `ClusterVirtualImages`: %s", err)
+			for _, cvi := range cviImgs.Items {
+				diskName := fmt.Sprintf("cvi-%s", cvi.Name)
+				imgs[diskName] = "value"
```
yaroslavborbat, line 360: Why is imgs[diskName] = "value" needed here? The value "value" is never actually used and gets overwritten later. I

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+		})
+
+		It("retrieves `VirtualMachine` object", func() {
+			vmObjs := &virtv2.VirtualMachineList{}
+			err := GetObjects(virtv2.VirtualMachineResource, vmObjs, kc.GetOptions{
+				Labels:    testCaseLabel,
+				Namespace: conf.Namespace,
+			})
+			Expect(err).NotTo(HaveOccurred(), "failed to get `VirtualMachines`: %s", err)
+			Expect(len(vmObjs.Items)).To(BeNumerically("==", vmCount), "there is only %d `VirtualMachine` in this test case", vmCount)
```
yaroslavborbat, line 278: HaveLen

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,442 @@
...
+}
+
+var _ = Describe("Image hotplug", func() {
+	var (
+		vmObj         virtv2.VirtualMachine
+		disksBefore   Disks
+		disksAfter    Disks
+		testCaseLabel = map[string]string{"testcase": "image-hotplug"}
+		isoLabel      = "iso"
+		blockDevices  = make([]Image, 0)
```
yaroslavborbat, line 113: It looks like blockDevices is shared across tests, which might lead to unintended dependencies between them. To keep tests isolated and predictable, consider using BeforeEach to initialize it locally within each test.

yaroslavborbat (changes requested):

tests/e2e/image_hotplug_test.go
hardcoretime: fixed

tests/e2e/image_hotplug_test.go
hardcoretime: It is part of a test where we check that the test case resources can be deleted correctly.

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: Unfortunately, we can't use BeforeAll because the prefix of all names is defined in the init function. If we use BeforeAll, the prefix value will not be defined in this section. It is known problem.

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: fixed

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: We can't use this variable in BeforeEach because we can retrieve images only after they have been created in one of the applied contexts.

tests/e2e/image_hotplug_test.go
yaroslavborbat: I still don't understand why it's necessary to first list images by label and then iterate through them using the VMI spec.

tests/e2e/image_hotplug_test.go
yaroslavborbat: Filling a global variable across different contexts and using it in another test is considered an anti-pattern and should be avoided. Although this might be part of an optimization strategy, it introduces dependencies between tests, which can lead to unpredictable behavior and make debugging harder.

I would recommend filling this variable before the tests start, in BeforeAll, and then cloning it before each test to ensure isolation. Alternatively, you can reinitialize the variable from scratch in each test to ensure that each test works with fresh, independent data. This would improve test reliability and maintainability.

yaroslavborbat (changes requested):

tests/e2e/image_hotplug_test.go
hardcoretime: This is not a cleanup. The Cleanup function is used before and after a test runs.
https://github.com/deckhouse/virtualization/blob/af7f31b82efa8c125591975e9e5e37be8ae4d986/tests/e2e/tests_suite_test.go#L147

tests/e2e/image_hotplug_test.go
hardcoretime: This does not work properly if a variable is defined outside the main `Describe`. However, you are right. We don't use global variables in the test cases anymore, and we should use `BeforeAll`.
```golang
var name = fmt.Sprintf("%s-name", namePrefix)

var _ = Describe("Image hotplug", ginkgoutil.CommonE2ETestDecorators(), func() {
	BeforeAll(func() {
		fmt.Printf("Show me the name: %s\n", name)
	})

	Context("test", func() {
		It("checks that `name` has the prefix", func() {
			Expect(name).To(HavePrefix(namePrefix))
		})
	})
})
```
```bash
$ FOCUS="Image hotplug" task e2e:run
  Starting test suite
Running Suite: Tests - ~/virtualization/tests/e2e
==========================================================================
Random Seed: 1739916614

Will run 1 of 151 specs
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
------------------------------
Image hotplug test checks that `name` has the prefix
~/virtualization/tests/e2e/image_hotplug_test.go:100
Show me the name: -name
  [FAILED] in [It] - ~/virtualization/tests/e2e/image_hotplug_test.go:101 @ 02/19/25 01:10:20.204
• [FAILED] [0.000 seconds]
Image hotplug test [It] checks that `name` has the prefix
~/virtualization/tests/e2e/image_hotplug_test.go:100

  [FAILED] Expected
      <string>: -name
  to have prefix
      <string>: head-af7f31b8
  In [It] at: ~/virtualization/tests/e2e/image_hotplug_test.go:101 @ 02/19/25 01:10:20.204
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Summarizing 1 Failure:
  [FAIL] Image hotplug test [It] checks that `name` has the prefix
  ~/virtualization/tests/e2e/image_hotplug_test.go:101

Ran 1 of 151 Specs in 0.002 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 150 Skipped
--- FAIL: TestTests (3.64s)
FAIL

Ginkgo ran 1 suite in 8.88048398s

Test Suite Failed
task: Failed to run task "e2e:run": exit status 1
```

tests/e2e/image_hotplug_test.go
hardcoretime: We need to obtain the images we want to test and their IDs from the VMI. This is because the VMI spec contains other types of disks that are not images.

tests/e2e/image_hotplug_test.go
hardcoretime: This variable cannot be assigned in `BeforeAll` or `BeforeEach` because we first need to apply the `Images` and only then store them in the variable. Additionally, this variable is used only in one `Context` where we attach `Images` to the `VirtualMachine`. I moved this variable to the container where it is used.

tests/e2e/image_hotplug_test.go
yaroslavborbat: made new comment

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,423 @@
...
+			Expect(err).NotTo(HaveOccurred(), "failed to get `InternalVirtulMachineInstance`: %s", err)
+			for _, disk := range intVirtVmi.Spec.Domain.Devices.Disks {
+				if _, ok := imgs[disk.Name]; ok {
+					if strings.HasSuffix(disk.Name, "iso") {
+						imgs[disk.Name] = fmt.Sprintf("%s-%s", CdRomIdPrefix, disk.Name)
+					} else {
+						imgs[disk.Name] = fmt.Sprintf("%s_%s", DiskIdPrefix, disk.Serial)
+					}
+				}
+			}
```
yaroslavborbat, line 369: Listing and populating the imgs map is still unnecessary. This results in extra API calls. Instead, you can directly check if disk.Name has the cvi- or vi- prefix while iterating over intVirtVmi.Spec.Domain.Devices.Disks and construct diskId on the fly.
```suggestion
			imgs := make(map[string]string, imgCount)
			intVirtVmi := &virtv1.VirtualMachineInstance{}
			err = GetObject(kc.ResourceKubevirtVMI, vmObj.Name, intVirtVmi, kc.GetOptions{
				Namespace: conf.Namespace,
			})
			Expect(err).NotTo(HaveOccurred(), "failed to get `InternalVirtulMachineInstance`: %s", err)
			for _, disk := range intVirtVmi.Spec.Domain.Devices.Disks {
				if strings.HasPrefix(disk.Name, "cvi-") || strings.HasPrefix(disk.Name, "vi-") {
					if strings.HasSuffix(disk.Name, "iso") {
						imgs[disk.Name] = fmt.Sprintf("%s-%s", CdRomIdPrefix, disk.Name)
					} else {
						imgs[disk.Name] = fmt.Sprintf("%s_%s", DiskIdPrefix, disk.Serial)
					}
				}
			}
```

tests/e2e/image_hotplug_test.go
yaroslavborbat: In this context, DeleteTestCaseResources is called—what is that if not resource cleanup?

In any case, even if this isn’t considered cleanup, if this context must always run last, then it should be in AfterAll.

tests/e2e/image_hotplug_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	BeforeAll(func() {
+		kustomization := fmt.Sprintf("%s/%s", conf.TestData.ImageHotplug, "kustomization.yaml")
+		ns, err := kustomize.GetNamespace(kustomization)
+		Expect(err).NotTo(HaveOccurred(), "%w", err)
+		conf.SetNamespace(ns)
+	})
+
+	Context("When the virtualization resources are applied", func() {
+		It("result should be succeeded", func() {
```
yaroslavborbat, line 153: I don't see the value of this BeforeAll. What exactly is being prepared here? All the setup still happens within the context, so nothing has really changed.

yaroslavborbat (changes requested):

tests/e2e/image_hotplug_test.go
yaroslavborbat: made new comment

tests/e2e/image_hotplug_test.go
hardcoretime: done

tests/e2e/image_hotplug_test.go
hardcoretime: If we want to refactor the current flow, I think it should be addressed in a separate task.

yaroslavborbat (approved):

PR #760, opened 2025-02-18
fix(vmbda): fix deletion vmbda already connected in vm spec

eofff:
## Description
Fixed a bug that prevented the deletion of the vmbda device when its block device was already specified in the virtual machine's specification.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vmbda
type: fix
summary: "fixed a bug that prevented the deletion of the vmbda device when its block device was already specified in the virtual machine's specification."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/service/attachment_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/attachment_service.go
@@ -151,7 +151,7 @@ func (s AttachmentService) HotPlugDisk(ctx context.Context, ad *AttachmentDisk, 
 }
 
 func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, vm *virtv2.VirtualMachine, blockDeviceName, originalName string, blockDeviceKind virtv2.VMBDAObjectRefKind) bool {
-	if blockDeviceName == "" || kvvm == nil || kvvm.Spec.Template == nil {
+	if blockDeviceName == "" || kvvm == nil || kvvm.Spec.Template == nil || vm == nil {
 		return false
 	}
 
--- a/images/virtualization-artifact/pkg/controller/service/attachment_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/attachment_service.go
@@ -156,12 +156,8 @@ func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, blockDeviceNam
 	}
 
 	for _, volume := range kvvm.Spec.Template.Spec.Volumes {
-		if !volume.PersistentVolumeClaim.Hotpluggable {
-			continue
-		}
-
 		if kvapi.VolumeExists(volume, blockDeviceName) {
-			return true
+			return volume.PersistentVolumeClaim.Hotpluggable
 		}
 	}
 
--- a/images/virtualization-artifact/pkg/controller/service/attachment_service.go
+++ b/images/virtualization-artifact/pkg/controller/service/attachment_service.go
@@ -150,19 +150,19 @@ func (s AttachmentService) HotPlugDisk(ctx context.Context, ad *AttachmentDisk, 
 	})
 }
 
-func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, vm *virtv2.VirtualMachine, diskName, originalName, diskKind string) bool {
-	if diskName == "" || kvvm == nil || kvvm.Spec.Template == nil {
+func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, vm *virtv2.VirtualMachine, blockDeviceName, originalName string, blockDeviceKind virtv2.VMBDAObjectRefKind) bool {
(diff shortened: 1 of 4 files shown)


Isteb4k (changes requested):
It's not clear what was wrong in the code and how the unplug method affected the removal of the failed vmbda

The description in the PR needs to be updated

images/virtualization-artifact/pkg/controller/service/attachment_service.go
```diff
@@ -150,11 +150,17 @@ func (s AttachmentService) HotPlugDisk(ctx context.Context, ad *AttachmentDisk,
 	})
 }
 
-func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, diskName string) bool {
+func (s AttachmentService) CanUnplug(kvvm *virtv1.VirtualMachine, vm *virtv2.VirtualMachine, diskName, originalName, diskKind string) bool {
 	if diskName == "" || kvvm == nil || kvvm.Spec.Template == nil {
 		return false
 	}
 
+	for _, specDisk := range vm.Spec.BlockDeviceRefs {
```
Isteb4k, line 158: See the status, not spec. The spec may be missing block devices that are nevertheless statically attached to the vm

images/virtualization-artifact/pkg/controller/vmbda/internal/deletion.go
```diff
@@ -89,7 +89,12 @@ func (h *DeletionHandler) cleanUp(ctx context.Context, vmbda *virtv2.VirtualMach
 		return err
 	}
 
-	if h.unplug.CanUnplug(kvvm, diskName) {
+	vm, err := object.FetchObject(ctx, types.NamespacedName{Namespace: vmbda.GetNamespace(), Name: vmbda.Spec.VirtualMachineName}, h.client, &virtv2.VirtualMachine{})
+	if err != nil {
+		return err
+	}
+
+	if h.unplug.CanUnplug(kvvm, vm, diskName, vmbda.Spec.BlockDeviceRef.Name, vmbda.Spec.BlockDeviceRef.Kind.String()) {
```
Isteb4k, line 97: Why did you decide to lose the type of kind and work with it as a string? Avoid doing that. Use VMBDAObjectRefKind.

images/virtualization-artifact/pkg/controller/vmbda/internal/deletion.go
```diff
@@ -89,7 +89,12 @@ func (h *DeletionHandler) cleanUp(ctx context.Context, vmbda *virtv2.VirtualMach
 		return err
 	}
 
-	if h.unplug.CanUnplug(kvvm, diskName) {
+	vm, err := object.FetchObject(ctx, types.NamespacedName{Namespace: vmbda.GetNamespace(), Name: vmbda.Spec.VirtualMachineName}, h.client, &virtv2.VirtualMachine{})
```
Isteb4k, line 92: It will panic below if the VM is nil

images/virtualization-artifact/pkg/controller/vmbda/internal/deletion.go
```diff
@@ -89,7 +89,12 @@ func (h *DeletionHandler) cleanUp(ctx context.Context, vmbda *virtv2.VirtualMach
...
+	if err != nil {
+		return err
+	}
+
+	if h.unplug.CanUnplug(kvvm, vm, diskName, vmbda.Spec.BlockDeviceRef.Name, vmbda.Spec.BlockDeviceRef.Kind.String()) {
 		h.log.Info("Unplug Virtual Disk", slog.String("diskName", diskName), slog.String("vm", kvvm.Name))
```
Isteb4k, line 98: Not only virtual disks, images as well

images/virtualization-artifact/pkg/controller/service/attachment_service.go
Isteb4k: Better add another check below to verify that this is a hot-pluggable disk (check Hotpluggable in virtv1.Volume) rather than looking at our virtual machine.

Next to VolumeExists, or better inside VolumeExists, but in that case, the function should be renamed.

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vmbda/internal/deletion.go
eofff: Because I cannot compare VMBDAObjectRefKind and BlockDeviceKind

eofff: Ok, now I am using string conversion only inside the function.

images/virtualization-artifact/pkg/controller/service/attachment_service.go
eofff: Now I am looking at the status)

Isteb4k (approved):

PR #785, opened 2025-02-21
feat(kubeviirt): remove unnecessary libvirt sockets

yaroslavborbat:
## Description  
Remove unnecessary libvirt sockets

Libvirt creates several sockets by default, but the following are not needed in our case:  

- `/var/run/libvirt/virtlogd-admin-sock`  
- `/var/run/libvirt/virtqemud-admin-sock`  
- `/var/run/libvirt/virtqemud-sock-ro`  

These sockets are removed after their creation to prevent any unintended interactions.  

The unnecessary sockets should no longer be available after libvirt initializes them, ensuring they do not interfere with our setup.  

```changes
section: kubevirt
type: feat
summary: remove unnecessary libvirt sockets
```

### DIFF
--- a/images/virt-artifact/patches/038-remove-unnecessary-libvirt-sockets.patch
+++ b/images/virt-artifact/patches/038-remove-unnecessary-libvirt-sockets.patch
@@ -1,0 +1,85 @@
+diff --git a/pkg/virt-launcher/virtwrap/util/file_cleaner.go b/pkg/virt-launcher/virtwrap/util/file_cleaner.go
+new file mode 100644
+index 0000000000..3d13818277
+--- /dev/null
++++ b/pkg/virt-launcher/virtwrap/util/file_cleaner.go
+@@ -0,0 +1,47 @@
++package util
++
++import (
++	"os"
++	"time"
++
++	"kubevirt.io/client-go/log"
++)
++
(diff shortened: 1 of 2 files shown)


images/virt-artifact/patches/038-remove-unnecessary-libvirt-sockets.patch
```diff
@@ -0,0 +1,85 @@
...
++	for {
++		select {
++		case <-stopChan:
++			return
++		default:
++			for f := range fileMap {
++				if err := os.Remove(f); err != nil {
++					if os.IsNotExist(err) {
++						continue
++					}
```
diafour, line 41: It seems non existent file should lead to `delete(fileMap, f)` but its not. The loop will wait forever if no socket created =)

yaroslavborbat: Yes, you are right. The thing is that these sockets do not exist initially. They are created by the child processes virtqemud and virtlogd. That’s why we need to wait for them to be created before deleting them.

diafour: 👍

PR #789, opened 2025-02-24
fix(core): distroless image fix bin,sbin directories

universal-itengineer:
## Description

All files in /bin,sbin,lib,lib64 directories should use symlinks to their respective counterparts in /usr

In modern Linux systems, the contents of /bin, /sbin, etc are often symlinked to their counterparts in /usr. 
This change simplifies system operation and improves maintainability

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: fix
summary: directories in root such as bin,sbin,etc linked to their respective counterparts in /usr
impact_level: low
```

### DIFF
--- a/images/distroless/werf.inc.yaml
+++ b/images/distroless/werf.inc.yaml
@@ -23,7 +23,13 @@ shell:
   - apt-get clean
   - rm --recursive --force /var/lib/apt/lists/ftp.altlinux.org* /var/cache/apt/*.bin
   install:
-  - mkdir -p /relocate/{usr,bin,sbin,etc} /relocate/etc/{pki,ssl} /relocate/usr/{bin,sbin,share}
+  - mkdir -p /relocate/etc/{pki,ssl} /relocate/usr/{bin,sbin,share,lib,lib64}
+  - |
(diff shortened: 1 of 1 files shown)


images/distroless/werf.inc.yaml
```diff
@@ -23,7 +23,13 @@ shell:
...
   - rm --recursive --force /var/lib/apt/lists/ftp.altlinux.org* /var/cache/apt/*.bin
   install:
-  - mkdir -p /relocate/{usr,bin,sbin,etc} /relocate/etc/{pki,ssl} /relocate/usr/{bin,sbin,share}
+  - mkdir -p /relocate/etc/{pki,ssl} /relocate/usr/{bin,sbin,share,lib,lib64}
+  - |
+    cd /relocate
+    for dir in {bin,sbin};do
+      ln -s usr/$dir $dir
+    done
+    cd /
```
diafour, line 32: Seems like 2 commands instead of 5, but OK.

diafour (approved):

nevermarine (approved):

PR #790, opened 2025-02-24
chore(core): distroless vm-route-forge

universal-itengineer:
## Description

Change vm-router-forge image to distroless

We want our images to be distroless-based

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: change vm-router-forge image to distroless
```

### DIFF
--- a/images/vm-route-forge/werf.inc.yaml
+++ b/images/vm-route-forge/werf.inc.yaml
@@ -28,16 +28,16 @@ shell:
   - export GOOS=linux
   - export GOARCH=amd64
   - export CGO_ENABLED=0
-  - go build -v -a -o vm-route-forge cmd/vm-route-forge/main.go
+  - go build -ldflags="-s -w" -o vm-route-forge cmd/vm-route-forge/main.go
 ---
 image: {{ $.ImageName }}
-from: {{ .Images.BASE_SCRATCH }}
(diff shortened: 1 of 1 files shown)


images/vm-route-forge/werf.inc.yaml
```diff
@@ -28,16 +28,16 @@ shell:
   - export GOOS=linux
   - export GOARCH=amd64
   - export CGO_ENABLED=0
-  - go build -v -a -o vm-route-forge cmd/vm-route-forge/main.go
+  - go build -ldflags="-s -w" -v -a -o vm-route-forge cmd/vm-route-forge/main.go
```
yaroslavborbat, line 31: why do you add ldflags `-s` and `-w`?

yaroslavborbat (changes requested):

images/vm-route-forge/werf.inc.yaml
universal-itengineer: -s it's strip binary
-w it's strips DWARF
all this params reduce the size of the binary

pprof should work

universal-itengineer:
pprof works correctly
image

yaroslavborbat (approved):

PR #796, opened 2025-02-25
fix(kubevirt): Ensure ContainerDisk is always mounted as Read-Only

yaroslavborbat:
## Description
Add patch `050-mount-containerdisk-always-readonly.patch`
This patch modifies the container's disk mounting behavior, ensuring the disk is always mounted in read-only mode.

This change is necessary to ensure a consistent approach in handling both CVI (Cluster Virtual Image) and VI (Virtual Image) resources. Currently, hot-plugged disks are mounted in read-only (RO) mode, while regular disks are mounted in read-write (RW) mode. This inconsistency can lead to unexpected behavior and complicates image management and usage.

By always mounting the container disk in read-only mode, we align the behavior across all disk types, improve security by preventing accidental or unintended modifications, and maintain a uniform handling of virtualization resources.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm 
type: fix
summary: Due to security issues, all images are mounted in 'read-only' mode.
```


diafour (approved):
lgtm

Isteb4k (approved):

PR #842, opened 2025-03-11
fix(core): revert scsi disk serial truncate in QEMU

yaroslavborbat:
## Description
Add patch `001-revert-scsi-disk-serial-truncate.patch`
This patch reverts the commit that introduced strict length enforcement for the SCSI disk `serial` property.

### Background

Before the reverted commit, scsi-disk accepted serial numbers of arbitrary length, but the value seen by the guest was silently truncated to 36 characters. While this limitation was arbitrary, it ensured compatibility with existing guest behavior. The change to enforce strict length validation introduced potential compatibility issues, making it impossible to upgrade to newer QEMU versions seamlessly.

### Reverted Commit
[Commit 75997e182b69](https://github.com/qemu/qemu/commit/75997e182b695f2e3f0a2d649734952af5caf3ee)

For the time being, we need to maintain backward compatibility until a seamless migration to the new behavior can be implemented. By reverting the commit, we restore the previous behavior where serial numbers longer than 36 characters are truncated instead of causing an error.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: core
type: fix
summary: Resolve potential compatibility issues related to the truncation of scsi disk serial numbers in QEMU.
```

### DIFF
--- a/images/qemu/patches/README.md
+++ b/images/qemu/patches/README.md
@@ -6,9 +6,7 @@ This patch reverts the commit that introduced strict length enforcement for the 
 
 ### Background
 
-Before the reverted commit, `scsi-disk` accepted serial numbers of arbitrary length, but the value seen by the guest was silently truncated to 36 characters. While this limitation was arbitrary, it ensured compatibility with existing guest behavior. The change to enforce strict length validation introduced potential compatibility issues by breaking the guest ABI for configurations that previously relied on truncation.
-
-### Why This Revert Is Necessary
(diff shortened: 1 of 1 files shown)


universal-itengineer (approved):

images/qemu/patches/001-revert-scsi-disk-serial-truncate.patch
```diff
@@ -0,0 +1,26 @@
...
+     if (!s->device_id) {
+         if (s->serial) {
+-            if (strlen(s->serial) > MAX_SERIAL_LEN_FOR_DEVID) {
+-                error_setg(errp, "The serial number can't be longer than %d "
+-                           "characters when it is also used as the default for "
+-                           "device_id", MAX_SERIAL_LEN_FOR_DEVID);
+-                return;
+-            }
+-            s->device_id = g_strdup(s->serial);
++	    s->device_id = g_strdup_printf("%.20s", s->serial);
```
diafour, line 23: README said serial is truncated to 36 and format string truncates to 20. Is it ok?

diafour (approved):

PR #852, opened 2025-03-12
fix(vi): validate `VirtualImage` storage class

hardcoretime:
## Description

A temporary solution until the problem with `PersistentVolumeMode` is resolved.

This is currently required to prevent a `VirtualImage` from being created with the `PersistentVolumeFilesystem` storage class.

 `VirtualImage` cannot be created the `PersistentVolumeFilesystem` storage class.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vi 
type: fix
summary:  "Prevent a `VirtualImage` from being created with the `PersistentVolumeFilesystem` storage class."
impact_level: high
impact: |
  `VirtualImage` cannot be created the `PersistentVolumeFilesystem` storage class.
```


images/virtualization-artifact/pkg/controller/vi/vi_webhook.go
```diff
@@ -114,3 +131,36 @@ func (v *Validator) ValidateDelete(_ context.Context, _ runtime.Object) (admissi
...
+			scProfile := &cdiv1.StorageProfile{}
+			err = v.client.Get(ctx, client.ObjectKey{Name: sc.Name}, scProfile, &client.GetOptions{})
+			if err != nil {
+				return admission.Warnings{}, fmt.Errorf("failed to obtain the `StorageProfile` %s", *scFromSpec)
+			}
+			if len(scProfile.Status.ClaimPropertySets) == 0 {
+				return admission.Warnings{}, fmt.Errorf("failed to validate the `PersistentVolumeMode` of the `StorageProfile`: %s", sc.Name)
+			}
+			if *scProfile.Status.ClaimPropertySets[0].VolumeMode == corev1.PersistentVolumeFilesystem {
+				return admission.Warnings{}, fmt.Errorf("a `StorageClass` with the `PersistentVolumeFilesystem` mode cannot be used currently")
```
Isteb4k, line 154: Suggest the message: "Currently, creating an image with storage on a PersistentVolumeClaim using the storage class %q is not supported, as it does not have the ability to use volume mode `Filesystem` and access mode `ReadWriteMany`. Please choose another storage class."

images/virtualization-artifact/pkg/controller/vi/vi_webhook.go
```diff
@@ -114,3 +131,36 @@ func (v *Validator) ValidateDelete(_ context.Context, _ runtime.Object) (admissi
...
+			err := v.client.Get(ctx, client.ObjectKey{Name: *scFromSpec}, sc, &client.GetOptions{})
+			if err != nil {
+				return admission.Warnings{}, fmt.Errorf("failed to obtain the `StorageClass` %s", *scFromSpec)
+			}
+			scProfile := &cdiv1.StorageProfile{}
+			err = v.client.Get(ctx, client.ObjectKey{Name: sc.Name}, scProfile, &client.GetOptions{})
+			if err != nil {
+				return admission.Warnings{}, fmt.Errorf("failed to obtain the `StorageProfile` %s", *scFromSpec)
+			}
+			if len(scProfile.Status.ClaimPropertySets) == 0 {
```
Isteb4k, line 150: It turns out that we don't take the first element from storageProfile.status.claimPropertySets, but rather [hardcode](https://github.com/deckhouse/virtualization/blob/main/images/virtualization-artifact/pkg/controller/service/disk_service.go#L179) the requirement for ReadWriteMany and Block. We need to check if the StorageProfile supports this combination.

images/virtualization-artifact/pkg/controller/vi/vi_webhook.go
```diff
@@ -114,3 +131,36 @@ func (v *Validator) ValidateDelete(_ context.Context, _ runtime.Object) (admissi
 	v.logger.Error("Ensure the correctness of ValidatingWebhookConfiguration", "err", err)
 	return nil, nil
 }
+
+func (v *Validator) validateStorageClass(ctx context.Context, vi *virtv2.VirtualImage) (admission.Warnings, error) {
```
Isteb4k, line 135: We want to check and prevent the creation of images not only for the storage class from the spec but also for default storage classes. This way, the user won't wait unnecessarily for an image that will never be created. Or was this check intentionally not added?

images/virtualization-artifact/pkg/controller/vi/internal/storageclass_ready.go
```diff
@@ -96,10 +96,24 @@ func (h StorageClassReadyHandler) Handle(ctx context.Context, vi *virtv2.Virtual
...
-			Reason(vicondition.StorageClassReady).
-			Message("")
+		scProfile, err := h.service.GetStorageProfile(ctx, sc.Name)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if len(scProfile.Status.ClaimPropertySets) == 0 {
+			return reconcile.Result{}, fmt.Errorf("failed to validate the `PersistentVolumeMode` of the `StorageProfile`: %s", sc.Name)
+		}
+		if *scProfile.Status.ClaimPropertySets[0].VolumeMode == corev1.PersistentVolumeFilesystem {
```
Isteb4k, line 106: Let's add the test case to ensure that the condition will be false if the storage class is not ReadWriteMany and Block.

images/virtualization-artifact/pkg/controller/moduleconfig/moduleconfig_webhook.go
```diff
@@ -54,5 +55,5 @@ func NewModuleConfigValidator(client client.Client) *validator.Validator[*mcapi.
 					oldMC.GetGeneration() != newMC.GetGeneration()
 			},
 		}).
-		WithUpdateValidators(cidrs, reduceCIDRs)
```
Isteb4k, line 57: What about WithCreateValidators ? Why not to validate using golang instead of python?

images/virtualization-artifact/pkg/controller/moduleconfig/util.go
```diff
@@ -91,3 +91,35 @@ func convertToStringSlice(input []interface{}) ([]string, error) {
...
+type viStorageClassSettings struct {
+	DefaultStorageClassName     string
+	AllowedStorageClassSelector AllowedStorageClassSelector
+}
+
+type AllowedStorageClassSelector struct {
+	MatchNames []string
+}
+
+func parseViStorageClass(settings mcapi.SettingsValues) *viStorageClassSettings {
```
Isteb4k, line 104: Why in util? It is only used in viStorageClassValidator. Let it stay there. We should strive to have less and less in util.

images/virtualization-artifact/pkg/controller/moduleconfig/moduleconfig_webhook.go
hardcoretime: This is required when the Virtualization-controller is absent from the cluster.

images/virtualization-artifact/pkg/controller/moduleconfig/util.go
hardcoretime: Done.

images/virtualization-artifact/pkg/controller/vi/vi_webhook.go
```diff
@@ -114,3 +131,36 @@ func (v *Validator) ValidateDelete(_ context.Context, _ runtime.Object) (admissi
...
+				return admission.Warnings{}, fmt.Errorf("failed to obtain the `StorageProfile` %s", *scFromSpec)
+			}
+			if len(scProfile.Status.ClaimPropertySets) == 0 {
+				return admission.Warnings{}, fmt.Errorf("failed to validate the `PersistentVolumeMode` of the `StorageProfile`: %s", sc.Name)
+			}
+			if *scProfile.Status.ClaimPropertySets[0].VolumeMode == corev1.PersistentVolumeFilesystem {
+				return admission.Warnings{}, fmt.Errorf("a `StorageClass` with the `PersistentVolumeFilesystem` mode cannot be used currently")
+			}
+		case virtv2.StorageKubernetes:
+			warnings := admission.Warnings{
```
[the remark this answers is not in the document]
hardcoretime: Done.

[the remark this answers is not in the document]
hardcoretime: Fixed.

images/virtualization-artifact/pkg/controller/vi/vi_webhook.go
hardcoretime: Done.

PR #857, opened 2025-03-17
refactor(vmop): impl new migration design

yaroslavborbat:
## Description
 Implement new migration design. Delete VMOP type Evict, equivalent to working as cancel migration.

```bash
 FOCUS="Virtual machine migration" task run
 Starting test suite

Ran 9 of 173 Specs in 130.829 seconds
SUCCESS! -- 9 Passed | 0 Failed | 0 Pending | 164 Skipped
PASS

```
## Checklist
- [X] The code is covered by unit tests.
- [X] e2e tests passed.

## Changelog entries

```changes
section: vmop
type: feature
summary: Implement new migration design. Delete VMOP type Evict, equivalent to working as cancel migration.
```


images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -137,6 +137,22 @@ func (h LifecycleHandler) Handle(ctx context.Context, s state.VMOperationState)
 			&changed.Status.Conditions)
 		return reconcile.Result{}, nil
 	}
+	// Fail if there is at least one other migration in progress.
+	found, err = h.vmopSrv.OtherMigrationsIsInProgress(ctx, changed)
```
Isteb4k, line 141: The function violates the rules of the English language.

Isteb4k, line 141: We don't need to call this for vmop start.

images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -137,6 +137,22 @@ func (h LifecycleHandler) Handle(ctx context.Context, s state.VMOperationState)
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if found {
+		changed.Status.Phase = virtv2.VMOPPhaseFailed
```
Isteb4k, line 146: Why did it fail? What does the user have to do with it?

The status is Failed, but the condition is as the user wants, 'InProgress'. What kind of error is this?

images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -182,12 +198,23 @@ func (h LifecycleHandler) syncOperationComplete(ctx context.Context, changed *vi
...
-	isComplete, err := h.vmopSrv.IsComplete(ctx, changed, vm)
+	isComplete, failureMessage, err := h.vmopSrv.IsComplete(ctx, changed, vm)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("check if operation is complete: %w", err)
 	}
 
 	if isComplete {
+		if failureMessage != "" {
+			changed.Status.Phase = virtv2.VMOPPhaseFailed
+			h.recorder.Event(changed, corev1.EventTypeNormal, virtv2.ReasonVMOPSucceeded, "VirtualMachineOperation failed")
```
Isteb4k, line 209: ReasonVMOPSucceeded: VirtualMachineOperation failed.

User: 😵‍💫

images/virtualization-artifact/pkg/controller/vmop/vmop_reconciler.go
```diff
@@ -94,6 +95,27 @@ func (r *Reconciler) SetupController(_ context.Context, mgr manager.Manager, ctr
 	); err != nil {
 		return fmt.Errorf("error setting watch on VirtualMachine: %w", err)
 	}
+
+	if err = ctr.Watch(
```
Isteb4k, line 99: Place it it internal/watcher

images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
```diff
@@ -182,12 +198,23 @@ func (h LifecycleHandler) syncOperationComplete(ctx context.Context, changed *vi
...
+			conditions.SetCondition(
+				completedCond.
+					Reason(vmopcondition.ReasonOperationFailed).
```
Isteb4k, line 212: Either failed or completed. It can't be a failed completed.
Completed is a successful completion.

api/client/builder/vmop.go
```diff
@@ -0,0 +1,74 @@
...
+
+	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
+)
+
+type VMOPBuilder struct {
+	*ObjectMetaBuilder
+	vmop *virtv2.VirtualMachineOperation
+}
+
+func NewVMOPBuilder(name, namespace string) *VMOPBuilder {
```
Isteb4k, line 30: Move from public api to internal

api/client/builder/metadata.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ObjectMetaBuilder struct {
+	objectMeta metav1.ObjectMeta
+}
+
+func NewObjectMetaBuilder(name, namespace string) *ObjectMetaBuilder {
```
Isteb4k, line 27: Move to internal as well

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vmop/internal/util.go
```diff
@@ -0,0 +1,23 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+import virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
+
+func isMigrate(vmop *virtv2.VirtualMachineOperation) bool {
```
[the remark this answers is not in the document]
yaroslavborbat: I initially planned to use this function in multiple handlers, which is why I moved it to a separate place. I agree that it can and should be relocated, but I still don’t quite understand what exactly makes this a "dangerous precedent." Are you referring to the potential growth of the util package, or is there another risk I’m missing?

images/virtualization-artifact/pkg/controller/vmop/internal/deletion.go
```diff
@@ -32,36 +34,62 @@ const deletionHandlerName = "DeletionHandler"
...
 
 func (h DeletionHandler) Handle(ctx context.Context, s state.VMOperationState) (reconcile.Result, error) {
-	log := logger.FromContext(ctx).With(logger.SlogHandler(deletionHandlerName))
+	h.log = logger.FromContext(ctx).With(logger.SlogHandler(deletionHandlerName))
 
 	if s.VirtualMachineOperation() == nil {
 		return reconcile.Result{}, nil
 	}
 
+	if isMigrate(s.VirtualMachineOperation().Changed()) {
```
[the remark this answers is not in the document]
yaroslavborbat: rename

images/virtualization-artifact/pkg/controller/vmop/internal/util.go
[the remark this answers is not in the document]
yaroslavborbat: moved

api/client/builder/metadata.go
yaroslavborbat: moved

images/virtualization-artifact/pkg/controller/vmop/internal/lifecycle.go
yaroslavborbat: fixed

images/virtualization-artifact/pkg/controller/vmop/internal/deletion.go
```diff
@@ -32,36 +34,62 @@ const deletionHandlerName = "DeletionHandler"
 
 // DeletionHandler manages finalizers on VirtualMachineOperation resource.
 type DeletionHandler struct {
-	client client.Client
+	client  client.Client
```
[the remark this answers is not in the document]
yaroslavborbat: What do you mean?

images/virtualization-artifact/pkg/controller/vmop/internal/deletion.go
```diff
@@ -32,36 +34,62 @@ const deletionHandlerName = "DeletionHandler"
...
+	vmopSrv service.VMOperationService
+	log     *slog.Logger
```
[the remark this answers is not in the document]
yaroslavborbat: I agree with the risks, but this was done intentionally since there's only one public method here, and it's responsible for initializing the logger upon invocation. This way, we don't have to constantly check for nil in private methods.

However, I think the bigger issue isn’t that we store the logger in the struct’s main method, but rather that we retrieve it from the context—where it might not always be present. This pattern is used throughout the project, meaning the entire approach isn’t fully safe.

To address this, I’ve updated the handler so that it always retrieves the logger from the context. Additionally, I modified the function that extracts the logger to include a nil check and return a default logger if none is available. This ensures that we always have a valid logger to work with.

Isteb4k (changes requested):

Isteb4k (approved):

PR #860, opened 2025-03-17
fix(core): use v1beta1 version for ValidatingAdmissionPolicy

diafour:
## Description

- Use v1beta1 so migration hook can work in Kubernetes 1.28 and 1.29.

Kubernetes subscriptions not support subscribe to nonexistent versions of CRD (yet).  ValidatingAdmissionPolicy/v1 exists in Kubernetes 1.30 and later, so we use v1beta1 for now. We can just delete this hook when v1beta1 will be removed (in 1.36 may be?).

Continuation of #839 

VAP Migration hook works in Kubernetes 1.28 and 1.29

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: fix
summary: Fix ValidatingAdmissionPolicy migration hook, it should support Kubernetes 1.28+
impact_level: low
```

### DIFF
--- a/hooks/migrate_delete_renamed_validation_admission_policy.py
+++ b/hooks/migrate_delete_renamed_validation_admission_policy.py
@@ -35,8 +35,6 @@ class MigrateDeleteRenamedVAPResources(Hook):
         self.managed_by_label_value = "virt-operator-internal-virtualization"
 
     def generate_config(self) -> dict:
-        kubernetesVersion = "1.30"
-
         return {
             "configVersion": "v1",
             "afterHelm": 5,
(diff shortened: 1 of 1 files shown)


hooks/migrate_delete_renamed_validation_admission_policy.py
```diff
@@ -35,13 +35,15 @@ def __init__(self, module_name: str):
         self.managed_by_label_value = "virt-operator-internal-virtualization"
 
     def generate_config(self) -> dict:
+        kubernetesVersion = "1.30"
+
```
diafour, line 39:
```suggestion
```

Isteb4k (approved):

PR #874, opened 2025-03-20
feat(vm): Add firmware handler 

yaroslavborbat:
## Description
Add firmware handler 

Add 2 new metric
- d8_virtualization_virtualmachine_info
- d8_virtualization_virtualmachine_firmware_up_to_date

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: feature
summary: Add firmware version in virtual machine status
```

### DIFF
--- a/api/core/v1alpha2/vmcondition/condition.go
+++ b/api/core/v1alpha2/vmcondition/condition.go
@@ -36,10 +36,10 @@ const (
 	TypeConfigurationApplied                Type = "ConfigurationApplied"
 	TypeAwaitingRestartToApplyConfiguration Type = "AwaitingRestartToApplyConfiguration"
 	// TypeFilesystemFrozen indicates whether the filesystem is currently frozen, a necessary condition for creating a snapshot.
-	TypeFilesystemFrozen    Type = "FilesystemFrozen"
-	TypeSizingPolicyMatched Type = "SizingPolicyMatched"
-	TypeSnapshotting        Type = "Snapshotting"
-	TypeFirmwareNeedUpdate  Type = "FirmwareNeedUpdate"
+	TypeFilesystemFrozen       Type = "FilesystemFrozen"
+	TypeSizingPolicyMatched    Type = "SizingPolicyMatched"
+	TypeSnapshotting           Type = "Snapshotting"
+	TypeFirmwareUpdateRequired Type = "FirmwareUpdateRequired"
 )
 
 type Reason string
@@ -110,5 +110,5 @@ const (
 	ReasonPodConditionMissing Reason = "PodConditionMissing"
 	ReasonGuestNotRunning     Reason = "GuestNotRunning"
 
-	ReasonFirmwareNeedUpdate Reason = "FirmwareNeedUpdate"
+	ReasonFirmwareUpdateRequired Reason = "FirmwareUpdateRequired"
 )
--- a/api/core/v1alpha2/vmcondition/condition.go
+++ b/api/core/v1alpha2/vmcondition/condition.go
@@ -39,6 +39,7 @@ const (
 	TypeFilesystemFrozen    Type = "FilesystemFrozen"
 	TypeSizingPolicyMatched Type = "SizingPolicyMatched"
 	TypeSnapshotting        Type = "Snapshotting"
+	TypeFirmwareNeedUpdate  Type = "FirmwareNeedUpdate"
 )
 
 type Reason string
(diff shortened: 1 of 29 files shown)


crds/virtualmachines.yaml
```diff
@@ -1241,6 +1241,9 @@ spec:
                         - size
                       type: object
                   type: object
+                firmwareVersion:
+                  type: string
+                  description: firmware Version.
```
z9r5, line 1246:
```suggestion
                  description: Firmware version.
```

z9r5 (dismissed):

images/virtualization-artifact/pkg/controller/vm/internal/firmware.go
```diff
@@ -0,0 +1,119 @@
...
+	changed := s.VirtualMachine().Changed()
+
+	kvvmi, err := s.KVVMI(ctx)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if kvvmi == nil || kvvmi.Status.LauncherContainerImageVersion == f.firmwareImage {
+		// If kvvmi does not exist, update the firmware version,
+		// as any newly created kvvmi will use the currently available firmware version.
```
Isteb4k, line 61: "Can we guarantee this? What if the virt-controller has been updated, but the virt-operator hasn't yet? In that case, the virtual machine will start with the old firmware, but the version will already be the new one.

images/virtualization-artifact/pkg/controller/vm/internal/firmware.go
```diff
@@ -0,0 +1,119 @@
...
+		return true
+	}
+	currVersion := version.Version(currentVersion)
+
+	if !currVersion.IsValid() {
+		return true
+	}
+
+	if f.firmwareVersion.Compare(currVersion) == 0 {
+		// Need update if versions is main but has different virt-launcher images
```
Isteb4k, line 91: It violates the rules of the English language. Check other comments as well

images/virtualization-artifact/pkg/version/version.go
```diff
@@ -16,6 +16,36 @@ limitations under the License.
 
 package version
 
-func GetEdition() string {
-	return edition
+import "golang.org/x/mod/semver"
+
+const MainVersion = "main"
```
Isteb4k, line 21: It's not clear why we need this const. Is it for debugging? Maybe it should be taken from firmware.yaml as well?

images/virtualization-artifact/pkg/version/firmware.yaml
```diff
@@ -0,0 +1,3 @@
+# Don't touch! Generating in CI
```
Isteb4k, line 1: There are two firmware.yaml files in the repo. Do we need both of them? If yes, add comments why

Isteb4k (changes requested):

Isteb4k (changes requested):
It's necessary to sync the PR with ADR.

api/core/v1alpha2/vmcondition/condition.go
```diff
@@ -39,6 +39,7 @@ const (
 	TypeFilesystemFrozen    Type = "FilesystemFrozen"
 	TypeSizingPolicyMatched Type = "SizingPolicyMatched"
 	TypeSnapshotting        Type = "Snapshotting"
+	TypeFirmwareNeedUpdate  Type = "FirmwareNeedUpdate"
```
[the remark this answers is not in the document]
yaroslavborbat: I replaced it with `FirmwareUpdateRequired`. Is that okay?
I don’t like how `FirmwareUpdateNeeded` sounds.

api/core/v1alpha2/vmcondition/condition.go
```diff
@@ -104,8 +105,10 @@ const (
...
-	ReasonPodTerminatingReason      Reason = "PodTerminating"
-	ReasonPodNotExistsReason        Reason = "PodNotExists"
-	ReasonPodConditionMissingReason Reason = "PodConditionMissing"
-	ReasonGuestNotRunningReason     Reason = "GuestNotRunning"
+	ReasonPodTerminating      Reason = "PodTerminating"
+	ReasonPodNotExists        Reason = "PodNotExists"
+	ReasonPodConditionMissing Reason = "PodConditionMissing"
+	ReasonGuestNotRunning     Reason = "GuestNotRunning"
+
+	ReasonFirmwareNeedUpdate Reason = "FirmwareNeedUpdate"
```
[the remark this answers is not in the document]
yaroslavborbat: FirmwareUpdateRequired

images/virtualization-artifact/pkg/version/version.go
yaroslavborbat: main is valid version for firmware. This const used in IsValid and IsMain methods in Version struct

images/virtualization-artifact/pkg/controller/vm/internal/firmware.go
yaroslavborbat: You're right. Currently, we cannot guarantee that our controller and KubeVirt are using the same configuration. A possible solution could be to introduce a special wait for the readiness of KubeVirt components during startup.

PR #902, opened 2025-04-02
fix(vi): handle vi default storage class from module config

hardcoretime:

When a `VirtualImage` is created without the `StorageClass` field in its specification and the default storage class is absent from the `ModuleConfig`, it is not handled after updating the default storage class in the `ModuleConfig`.

The default storage class field works properly in the reconciliation loop of the `VirtualImage` resource.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vi 
type: fix
summary: "The `VirtualImageDefaultStorageClass` from the `ModuleConfig` is handled correctly now." 
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/indexer/vi_indexer.go
+++ b/images/virtualization-artifact/pkg/controller/indexer/vi_indexer.go
@@ -19,10 +19,12 @@ package indexer
 import (
 	"context"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	virtv2 "github.com/deckhouse/virtualization/api/core/v1alpha2"
+	"github.com/deckhouse/virtualization/api/core/v1alpha2/vicondition"
 )
 
 func IndexVIByVDSnapshot(ctx context.Context, mgr manager.Manager) error {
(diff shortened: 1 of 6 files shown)


images/virtualization-artifact/pkg/controller/indexer/vi_indexer.go
```diff
@@ -61,3 +63,20 @@ func IndexVIByStorageClass(ctx context.Context, mgr manager.Manager) error {
 		}
 	})
 }
+
+func IndexVIByNotReadyStorageClass(ctx context.Context, mgr manager.Manager) error {
```
eofff, line 67: Why with this name you indexing by sc ready? And ignore non pvc VI please.

images/virtualization-artifact/pkg/controller/service/disk_service.go
```diff
@@ -608,8 +609,32 @@ func (s DiskService) GetStorageClass(ctx context.Context, storageClassName *stri
...
+		moduleConfigViDefaultStorageClass string
+		moduleConfig                      mcapi.ModuleConfig
+		moduleConfigName                  = "virtualization"
+	)
+	err := s.client.Get(ctx, types.NamespacedName{Name: moduleConfigName}, &moduleConfig, &client.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	if virtualImages, ok := moduleConfig.Spec.Settings["virtualImages"].(map[string]interface{}); ok {
```
eofff, line 622: Do not hardcode please, use constant not string literal (set it if needed), next line too

images/virtualization-artifact/pkg/controller/service/disk_service.go
```diff
@@ -608,8 +609,32 @@ func (s DiskService) GetStorageClass(ctx context.Context, storageClassName *stri
 }
 
 func (s DiskService) GetDefaultStorageClass(ctx context.Context) (*storev1.StorageClass, error) {
+	var (
+		moduleConfigViDefaultStorageClass string
+		moduleConfig                      mcapi.ModuleConfig
+		moduleConfigName                  = "virtualization"
```
eofff, line 615: Can it change?

images/virtualization-artifact/pkg/controller/service/disk_service.go
eofff: "virtualImages" var seems like slice of VI, use virtualImagesSetting or something

images/virtualization-artifact/pkg/controller/vi/internal/watcher/moduleconfig_watcher.go
```diff
@@ -0,0 +1,107 @@
...
+				oldMc, oldOk := event.ObjectOld.(*mcapi.ModuleConfig)
+				newMc, newOk := event.ObjectNew.(*mcapi.ModuleConfig)
+				if !oldOk || !newOk {
+					return false
+				}
+				var (
+					oldViDefaultSc string
+					newViDefaultSc string
+				)
+				if virtualImages, ok := oldMc.Spec.Settings["virtualImages"].(map[string]interface{}); ok {
```
eofff, line 70: Use consts, not string literals

eofff (changes requested):

PR #929, opened 2025-04-14
fix(vi,vd): fix default storage class usage

hardcoretime:
## Description

- handle default storage class from the module config
- fix `StorageClassReady` condition

When a VirtualImage is created without the StorageClass field in its specification and the default storage class is absent from the ModuleConfig, it is not handled after updating the default storage class in the ModuleConfig.

The default storage class field works properly in the reconciliation loop of the VirtualImage resource.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vi,vd 
type: fix
summary: "The `VirtualImageDefaultStorageClass` from the `ModuleConfig` is handled correctly now." 
```

### DIFF
--- a/api/core/v1alpha2/vicondition/condition.go
+++ b/api/core/v1alpha2/vicondition/condition.go
@@ -79,8 +79,6 @@ const (
 	ProvisioningNotStarted ReadyReason = "ProvisioningNotStarted"
 	// ProvisioningFailed indicates that the provisioning process has failed.
 	ProvisioningFailed ReadyReason = "ProvisioningFailed"
-	// StorageClassNotReady indicates that the provisioning process pending because `StorageClass` not ready.
-	StorageClassNotReady ReadyReason = "StorageClassNotReady"
 	// Ready indicates that the import process is complete and the `VirtualImage` is ready for use.
 	Ready ReadyReason = "Ready"
 	// QuotaExceeded indicates that the VirtualImage is reached project quotas and can not be provisioned.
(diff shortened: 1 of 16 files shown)


api/core/v1alpha2/vicondition/condition.go
```diff
@@ -93,6 +91,8 @@ const (
 
 	// StorageClassReady indicates that the chosen StorageClass exists.
 	StorageClassReady StorageClassReadyReason = "StorageClassReady"
+	// StorageClassNotReady indicates that the provisioning process pending because `StorageClass` not ready.
+	StorageClassNotReady ReadyReason = "StorageClassNotReady"
```
Isteb4k, line 95: Note that the reasons are grouped by type (ReadyReason is placed next to ReadyReason, StorageClassReadyReason is placed next to StorageClassReadyReason). But here, a ReadyReason has ended up among the StorageClassReadyReasons, even though it should be grouped with its own kind

images/virtualization-artifact/pkg/controller/vi/internal/storageclass_ready.go
```diff
@@ -71,58 +72,162 @@ func (h StorageClassReadyHandler) Handle(ctx context.Context, vi *virtv2.Virtual
 	}
```
Isteb4k, line 72: Please add `return reconcile.Result{}, nil` for when DeletionTimestamp != nil, for line 59

images/virtualization-artifact/pkg/controller/service/vi_storage_class_service.go
```diff
@@ -47,7 +52,7 @@ func NewVirtualImageStorageClassService(settings config.VirtualImageStorageClass
 // Errors:
 // 1. Return error if no storage class is specified.
 // 2. Return error if specified non-empty class is not allowed.
-func (svc *VirtualImageStorageClassService) GetStorageClass(storageClassFromSpec *string, clusterDefaultStorageClass *storev1.StorageClass) (*string, error) {
+func (svc *VirtualImageStorageClassService) GetValidatedStorageClass(storageClassFromSpec *string, clusterDefaultStorageClass *storev1.StorageClass) (*string, error) {
```
Isteb4k, line 55: Up to date with the new name the comment above.

Isteb4k (changes requested):

PR #933, opened 2025-04-15
test(vmop): add test vmop migration cancel

yaroslavborbat:
## Description
add test vmop migration cancel

## Checklist
- [X] e2e tests passed.

## Changelog entries

```changes
section: vmop
type: feature
summary: add test vmop migration cancel 
impact_level: low
```

### DIFF
--- a/tests/e2e/testdata/vm-migration-cancel/kustomization.yaml
+++ b/tests/e2e/testdata/vm-migration-cancel/kustomization.yaml
@@ -1,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: testcases
+namePrefix: pr-number-or-commit-hash-
+resources:
+  - vi
+  - vm
+  - ns.yaml
+configurations:
+  - transformer.yaml
+labels:
+  - includeSelectors: true
+    pairs:
+      id: pr-number-or-commit-hash
+      testcase: vm-migration-cancel
--- a/tests/e2e/util_test.go
+++ b/tests/e2e/util_test.go
@@ -297,7 +297,10 @@ func WaitByLabel(resource kc.Resource, opts kc.WaitOptions) {
 	})
 	Expect(res.Error()).NotTo(HaveOccurred(), res.StdErr())
 
-	resources := strings.Split(res.StdOut(), " ")
+	var resources []string
+	if stdout := res.StdOut(); stdout != "" {
+		resources = strings.Split(res.StdOut(), " ")
+	}
 	WaitResources(resources, resource, opts)
 }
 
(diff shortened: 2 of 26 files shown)


danilrwx (dismissed):

danilrwx (dismissed):

tests/e2e/util_test.go
```diff
@@ -665,3 +668,15 @@ func SaveTestResources(labels map[string]string, additional string) {
 	err := os.WriteFile(str, cmdr.StdOutBytes(), 0644)
 	Expect(err).NotTo(HaveOccurred())
 }
+
+func ListResource(resource kc.Resource, namespace string, matchLabels map[string]string, out client.ObjectList) error {
```
hardcoretime, line 672: There is `GetObjects` function: https://github.com/deckhouse/virtualization/blob/285453881ca0fcfd9e627c64a113e20759ff2024/tests/e2e/util_test.go#L224

tests/e2e/vm_migration_test.go
```diff
@@ -34,9 +37,9 @@ const (
 	KubevirtVMIMKind   = "InternalVirtualizationVirtualMachineInstanceMigration"
```
hardcoretime, line 37: Can be deleted.

tests/e2e/testdata/vm-migration-cancel/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
hardcoretime, line 1: Is it necessary to create a full template directory? I think we can add the required `VirtualMachines` to `testdata/vm-migration/`.

tests/e2e/vm_migration_test.go
```diff
@@ -80,28 +83,6 @@ var _ = Describe("Virtual machine migration", ginkgoutil.CommonE2ETestDecorators
 		})
 	})
 
-	Context("When virtual images are applied", func() {
```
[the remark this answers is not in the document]
yaroslavborbat: I understand your concern about preserving detailed reports and ensuring that all resources are accounted for. However, I believe that simplicity and focus in our tests are crucial for maintainability and clarity.

The primary goal of these tests is to validate the VM functionality, not to verify secondary resources. Those secondary resources already have their own dedicated tests, and we should rely on those tests to ensure their correctness. Including additional checks here only adds unnecessary complexity and increases the risk of unrelated failures obscuring the actual purpose of the test.

By focusing solely on the VM, we make the tests easier to read, maintain, and debug. This approach aligns with the principle of separation of concerns, where each test suite has a specific responsibility. If there are any issues with secondary resources, they will be caught in their respective test suites, and we can address them there.

hardcoretime (approved):

PR #947, opened 2025-04-18
fix(vd,vi): fix issues with storage class condition display

Isteb4k:
## Description

1. The `Start` and `StartImmediate` methods of `DiskService` now take a `*storagev1.StorageClass` instead of a `string` as the storage class parameter.

2. The global services `VirtualImageStorageClassService` and `VirtualDiskStorageClassService` have been moved to the _internal/service_ package of the corresponding resource.

3. All checks related to `StorageClass` are now performed exclusively in the `StorageClassReady` handlers. The `LifeCycle` handler (and all source handlers) now rely on the `StorageClassReady` condition and no longer perform redundant checks.

4. A new service, `BaseStorageClassService`, which integrates internal storage class services in order to use the base logic for working with `StorageClass`.

5. Refactored `LifeCycle` for vi similarly to how it’s done for vd.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: api 
type: fix
summary: fix issues with storage class condition display
```

### DIFF
--- a/api/core/v1alpha2/events.go
+++ b/api/core/v1alpha2/events.go
@@ -64,12 +64,11 @@ const (
 	// ReasonVMOPInProgress is event reason that the operation is in progress
 	ReasonVMOPInProgress = "VirtualMachineOperationInProgress"
 
-	// ReasonVDStorageClassWasDeleted is event reason that VDStorageClass was deleted.
-	ReasonVDStorageClassWasDeleted = "VirtualDiskStorageClassWasDeleted"
-	// ReasonVDStorageClassNotFound is event reason that VDStorageClass not found.
-	ReasonVDStorageClassNotFound = "VirtualDiskStorageClassNotFound"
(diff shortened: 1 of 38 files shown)


hardcoretime:
This isn't about your changes, but I think we can refactor this [section](https://github.com/deckhouse/virtualization/blob/0edd300f6d7f9b2f899a74d24a9359c422bae50e/images/virtualization-artifact/pkg/controller/vi/internal/storageclass_ready.go#L97) in this PR. When we discussed this section with @yaroslavborbat in another PR, he pointed out that we don't need to get the default storage class again because `GetModuleStorageClass` [returns](https://github.com/deckhouse/virtualization/blob/main/images/virtualization-artifact/pkg/controller/service/disk_service.go#L605) the default storage class or `ErrDefaultStorageClassNotFound` if it does not exist.

hardcoretime (approved):

PR #890, opened 2025-03-28
feat(vm): live migration policies

diafour:
## Description

- Add per-vm migration policies.
- Add kubevirt patch to handoff MigrationConfiguration construction to the virtualization-controller.

## Why do we need it, and what problem does it solve?

- Set autoConverge option via VM and VMOP.

Alternative solution: kubevirt MigrationPolicy resources. But we need more simpler and more dynamic solution:
- MigrationPolicy should be created before creating migration.
- MigrationPolicy resources should be recreated to change options.

Dynamic settings seems more perspective, so Kubevirt was patched to handoff constructing of MigrationConfiguration to the virtualization-controller.

## What is the expected result?

- User can set AlwaysForced policy and VM will always migrate with autoConverge.
- User can migrate with enable/disable autoConverge via VMOP.

## Checklist
- [x] The code is covered by unit tests.
- [x] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: vm
type: feature
summary: Ability to force migration with CPU throttling. Live migration policy can be set in VM and user can override its value with VMOP.
```

PR #964, opened 2025-04-22
chore(vm): change virt-launcher for testing pr890

diafour:
## Description

Temporary PR for testing #890 .


diafour:
Not needed anymore

PR #981, opened 2025-04-28
test(vdsnapshot): disable filesystem frozen check

eofff:
## Description

It is a known issue that filesystem not always freeze while vd snapshot tests.
To prevent this error from causing noise during testing, we disabled this check.
It will need to be re-enabled once this issue is fixed.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vdsnapshot
type: chore
summary: disable filesystem frozen check
impact_level: low
```


tests/e2e/vd_snapshots_test.go
```diff
@@ -384,8 +387,47 @@ var _ = Describe("Virtual disk snapshots", ginkgoutil.CommonE2ETestDecorators(),
...
+
+			for _, vm := range vms.Items {
+				frozenCondition, _ := conditions.GetCondition(vmcondition.TypeFilesystemFrozen, vm.Status.Conditions)
+				vmsFrozenDict[vm.Name] = frozenCondition.Status == v1.ConditionTrue
+			}
+
+			watcher, err := virtClient.VirtualMachines(conf.Namespace).Watch(context.Background(), v1.ListOptions{TimeoutSeconds: ptr.To(int64(filesystemReadyTimeout.Seconds()))})
+			Expect(err).NotTo(HaveOccurred())
+			var hasFrozen bool
+			for {
```
yaroslavborbat, line 402: need timeout

yaroslavborbat:
```go
for {
  select {
  case <-time.After(5 * time.Minute):
    panic ....
  }
}
```

tests/e2e/vd_snapshots_test.go
```diff
@@ -384,8 +387,47 @@ var _ = Describe("Virtual disk snapshots", ginkgoutil.CommonE2ETestDecorators(),
...
+			vmsFrozenDict := make(map[string]bool)
+
+			for _, vm := range vms.Items {
+				frozenCondition, _ := conditions.GetCondition(vmcondition.TypeFilesystemFrozen, vm.Status.Conditions)
+				vmsFrozenDict[vm.Name] = frozenCondition.Status == v1.ConditionTrue
+			}
+
+			watcher, err := virtClient.VirtualMachines(conf.Namespace).Watch(context.Background(), v1.ListOptions{TimeoutSeconds: ptr.To(int64(filesystemReadyTimeout.Seconds()))})
+			Expect(err).NotTo(HaveOccurred())
+			var hasFrozen bool
```
yaroslavborbat, line 401: declare in for

tests/e2e/vd_snapshots_test.go
```diff
@@ -384,8 +387,47 @@ var _ = Describe("Virtual disk snapshots", ginkgoutil.CommonE2ETestDecorators(),
...
+					watcher.Stop()
+					break
+				}
+
+				event, ok := <-watcher.ResultChan()
+				if !ok {
+					break
+				}
+
+				vm, ok := event.Object.(*virtv2.VirtualMachine)
```
yaroslavborbat, line 418: Is it old or new object?

yaroslavborbat (changes requested):

tests/e2e/vd_snapshots_test.go
eofff: We already have timeout in ListOptions

tests/e2e/vd_snapshots_test.go
eofff: https://github.com/kubernetes/apimachinery/blob/1bb0bf9dfa7eb27a21dd17b7cabedb876ec5a835/pkg/watch/watch.go#L74C2-L74C66

tests/e2e/vd_snapshots_test.go
eofff: After fix problem, we will use it state outer for

tests/e2e/vd_snapshots_test.go
Isteb4k, line 402: Let's create a generic solution that can be reused in other tests, so we don't have to write these kind of for loops each time.

Isteb4k (changes requested):

hardcoretime:
I think we should uncomment this check because the test does not check that the snapshot becomes ready.
https://github.com/deckhouse/virtualization/blob/48bdf4886c9a1d6ddd07a2a06dd885f510ab2cce/tests/e2e/vd_snapshots_test.go#L545

hardcoretime (changes requested):
I think we should uncomment this check because the test does not check that the snapshot becomes ready.
https://github.com/deckhouse/virtualization/blob/48bdf4886c9a1d6ddd07a2a06dd885f510ab2cce/tests/e2e/vd_snapshots_test.go#L545

deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/15207122448)

hardcoretime (approved):

tests/e2e/vd_snapshots_test.go
```diff
@@ -222,6 +225,53 @@ func CheckFileSystemFrozen(vmName string) (bool, error) {
...
+	watcher, err := virtClient.VirtualMachines(conf.Namespace).Watch(context.Background(), v1.ListOptions{TimeoutSeconds: ptr.To(timeoutSeconds)})
+	if err != nil {
+		return false, err
+	}
+
+	var hasFrozen bool
+	for {
+		hasFrozen = false
+		for _, frozen := range vmFSFrozenByName {
+			hasFrozen = hasFrozen || frozen
```
Isteb4k, line 250: hasFrozen is always false here; you unconditionally set it to false above

Isteb4k: Why do you need to iterate all the virtual machines if you've already found the frozen one? Why not just use a break for the for loop at line 249?

tests/e2e/vd_snapshots_test.go
```diff
@@ -222,6 +225,53 @@ func CheckFileSystemFrozen(vmName string) (bool, error) {
...
+		}
+
+		event, ok := <-watcher.ResultChan()
+		if !ok {
+			break
+		}
+
+		vm, ok := event.Object.(*virtv2.VirtualMachine)
+		if !ok {
+			continue
```
Isteb4k, line 265: This will result in an infinite loop in case of an error. If something goes wrong, throw an error in the test

Isteb4k (changes requested):

PR #1057, opened 2025-05-16
chore(ci): build libblkid

LopatinDmitr:
## Description

Build libblkid library.

We want more control when building libraries files, as well as making images more secure

All work as expected

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: build libblkid
impact_level: low
```

### DIFF
--- a/images/packages/binaries/libblkid/werf.inc.yaml
+++ b/images/packages/binaries/libblkid/werf.inc.yaml
@@ -45,10 +45,7 @@ shell:
     git clone --depth=1 $(cat /run/secrets/SOURCE_REPO)/{{ $gitRepoUrl }} --branch {{ $version }} /src
     cd /src
     ./autogen.sh
-    mkdir ../build
-    cd ../build
-
-    ../src/configure \
+    ./configure \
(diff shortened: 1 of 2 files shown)


images/packages/binaries/libblkid/werf.inc.yaml
```diff
@@ -0,0 +1,61 @@
...
+      mkdir -p ~/.ssh && echo "StrictHostKeyChecking accept-new" > ~/.ssh/config
+
+      git clone --depth=1 $(cat /run/secrets/SOURCE_REPO)/{{ $gitRepoUrl }} --branch {{ $version }} /src
+      cd /src
+      ./autogen.sh
+      cd ../
+      mkdir /build
+      cd /build
+
+      ../src/configure \
```
diafour, line 50: Something unusual, why not run from /src?

LopatinDmitr: Because the files we obtain as a result of running the program are needed in the build folder.

PR #1061, opened 2025-05-16
chore(vm): add label to prevent node shutdown, remove kubectl anno from kvvmi template

diafour:
## Description

- Add label pod.deckhouse.io/inhibit-node-shutdown to support upcoming deckhouse feature to prevent node shutdown
- Remove SetMetadata from ApplyVirtualMachineSpec to remove kubectl.kubernetes.io annotation from kvvmi template in new kvvm

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: chore
summary: Add node shutdown inhibit label to VM Pod, remove kubectl.kubernetes.io annotation from KVVMI.
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
+++ b/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
@@ -587,18 +587,3 @@ func (b *KVVM) GetBootloaderSettings() map[string]interface{} {
 		},
 	}
 }
-
-func (b *KVVM) SetMetadata(metadata metav1.ObjectMeta) {
-	if b.ResourceExists {
-		// initialize only
-		return
-	}
-	if b.Resource.Spec.Template.ObjectMeta.Labels == nil {
-		b.Resource.Spec.Template.ObjectMeta.Labels = make(map[string]string, len(metadata.Labels))
-	}
-	if b.Resource.Spec.Template.ObjectMeta.Annotations == nil {
-		b.Resource.Spec.Template.ObjectMeta.Annotations = make(map[string]string, len(metadata.Annotations))
-	}
-	maps.Copy(b.Resource.Spec.Template.ObjectMeta.Labels, metadata.Labels)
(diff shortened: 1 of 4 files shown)


images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
```diff
@@ -587,18 +587,3 @@ func (b *KVVM) GetBootloaderSettings() map[string]interface{} {
...
-	}
-	if b.Resource.Spec.Template.ObjectMeta.Labels == nil {
-		b.Resource.Spec.Template.ObjectMeta.Labels = make(map[string]string, len(metadata.Labels))
-	}
-	if b.Resource.Spec.Template.ObjectMeta.Annotations == nil {
-		b.Resource.Spec.Template.ObjectMeta.Annotations = make(map[string]string, len(metadata.Annotations))
-	}
-	maps.Copy(b.Resource.Spec.Template.ObjectMeta.Labels, metadata.Labels)
-	maps.Copy(b.Resource.Spec.Template.ObjectMeta.Annotations, metadata.Annotations)
-}
```
yaroslavborbat, line 604: We cannot remove this method because we rely on it to set annotations and labels when creating a resource.
Our reconciler that synchronizes metadata between resources works well, but has limitations — specifically, it works with a delay and updates only existing resources.
For example, affinity and anti-affinity rules will stop working correctly if metadata is not set at creation time.

Please keep the method to ensure metadata is properly initialized during resource creation.

yaroslavborbat (changes requested):

images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
diafour: I've reverted this deletion, will make another PR to fix kubectl annotation propagation.

yaroslavborbat (approved):

PR #1101, opened 2025-05-28
chore(ci): build lvm2

eofff:
## Description
Build lvm2 because we need libdevmapper.

We want more control when building libraries files, as well as making images more secure.

All work as expected.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: build lvm2
impact_level: low
```


images/packages/binaries/lvm2/werf.inc.yaml
```diff
@@ -0,0 +1,66 @@
...
+final: false
+fromImage: builder/scratch
+import:
+- image: {{ $.ImageType }}/{{ $.ImageName }}-builder
+  add: /out
+  to: /libdevmapper
+  before: setup
+
+---
+{{- $version := "v2_03_32" }}
```
universal-itengineer, line 12: Use this to get the version from a file `component_versions/version_map.yml`
```suggestion
{{- $version := get $.Package $.ImageName }}
```

PR #1112, opened 2025-05-30
chore(ci): build libgcc1

eofff:
## Description
Build gcc because we need libgcc1 library.

We want more control when building libraries files, as well as making images more secure

All work as expected

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: build gcc
impact_level: low
```

### DIFF
--- a/images/packages/binaries/gcc/werf.inc.yaml
+++ b/images/packages/binaries/gcc/werf.inc.yaml
@@ -73,6 +73,7 @@ shell:
     export ac_cv_func_fprintf_unlocked="no"
 
     ./configure \
+      --enable-languages=c \
       --prefix=/usr \
       --libdir=/usr/lib64 \
       --enable-shared \


images/packages/binaries/gcc/werf.inc.yaml
```diff
@@ -0,0 +1,98 @@
...
+final: false
+fromImage: builder/scratch
+import:
+- image: {{ $.ImageType }}/{{ $.ImageName }}-builder
+  add: /out
+  to: /gcc
+  before: setup
+
+---
+{{- $version := "releases/gcc-14.2.0" }}
```
universal-itengineer, line 12: Use this to get the version from a file `component_versions/version_map.yml`
```suggestion
{{- $version := get $.Package $.ImageName }}
```

images/packages/binaries/gcc/werf.inc.yaml
```diff
@@ -0,0 +1,98 @@
...
+            d="${f%%/*}"; \
+            grep '^m4_include' "$d/aclocal.m4" \
+              | grep -E -v '\[(libltdl/)?acinclude\.m4\]' \
+              > acinclude.m4~ || :; \
+            touch "$d/acinclude.m4"; \
+            cat  "$d/acinclude.m4" >> acinclude.m4~; \
+            mv   acinclude.m4~ "$d/acinclude.m4"; \
+            autoreconf -fvi "$d"; \
+            sh -n "$d/configure"; \
+        done
```
diafour, line 68: Please, add comment describing the purpose of this spell =)

PR #1134, opened 2025-06-06
fix(module): will not deploy virtualization if moduleconfig is not valid

yaroslavborbat:
## Description

Set isReady flag if required fields in ModuleConfig are valid (i.e. cidrs).

User should be able to enable virtualization module with empty settings in ModuleConfig.

There are no mechanics in deckhouse to prevent module enablement with empty settings, so we use a workaround: do not deploy module resources if ModuleConfig is not valid.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: fix
summary:  Do not deploy virtualization module resources if ModuleConfig is not valid when module is enabled at first time.
```

### DIFF
--- a/hooks/module_config_validator.py
+++ b/hooks/module_config_validator.py
@@ -30,7 +30,7 @@ class ModuleConfigValidateHook(Hook):
     def __init__(self, module_name: str):
         self.module_name = module_name
         self.queue = f"/modules/{self.module_name}/{self.SNAPSHOT_MODULE_CONFIG}"
-        self.path = "virtualization.internal.ready"
+        self.path = "virtualization.internal.isReady"
 
     def generate_config(self) -> dict:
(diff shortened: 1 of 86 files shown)


hooks/module_config_validator.py
```diff
@@ -86,16 +87,28 @@ def r(ctx: hook.Context):
...
                 .get("cidrs", [])
             ]
-            self.check_overlaps_cidrs(cidrs)
+
+            try:
+                self.check_overlaps_cidrs(cidrs)
+            except ValueError as e:
+                print(f"ERROR: {e}")
+
+                self.set_value(self.path, ctx.values, False)
```
diafour, line 96: Will it ends up here if there are no "cidrs" in ModuleConfig?
It seems there should be another check to set isReady=False if "cidrs" are empty.

PR #1161, opened 2025-06-16
chore(ci): build libidn

LopatinDmitr:
## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: 
type: 
summary: 
```


images/packages/binaries/libidn/werf.inc.yaml
```diff
@@ -0,0 +1,71 @@
...
+  before: setup
+
+---
+{{- $version := get $.Package $.ImageName }}
+{{- $gitRepoUrl := "libidn/libidn.git" }}
+
+{{- $name := print $.ImageName "-dependencies" -}}
+{{- define "$name" -}}
+packages:
+- gcc git make libtool wget patch gperf help2man gengetopt
```
diafour, line 31: Can you check why wget is needed? It is not good if make or configure use wget to download some assets or sources, we need build to be isolatable from the web.

LopatinDmitr: Fixed

diafour (approved):

PR #859, opened 2025-03-17
feat(images): add VI/CVI InUse condition

eofff:
## Description

- Add new condition InUse
- Add new handler to set InUse condition

## Why do we need it, and what problem does it solve?

Other resources may use CVI/VI, we want to display which resources are blocking image deletion.

## What is the expected result?

User can see which resources to check to unblock image deletion.

## Checklist
- [x] The code is covered by unit tests.
- [x] e2e tests passed.
- [x] Documentation updated according to the changes.
- [x] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: images
type: feature
summary: "Add VI/CVI InUse condtion"
```

PR #1165, opened 2025-06-16
chore(vd): fix phase name for PVCLost phase

diafour:
## Description

- Sync go source for VD: Use "PVCLost" as phase name for DiskLost phase.

Fixes for "CRD only changes":
- Sync go source for VD: change jsonpath for StorageClass column.
- Sync go source for CVI: add description to UsedInNamespaces fields.

Disk deletion is stuck after changes in #859. Phase was not used before, so we don't catch this problem earlier.

- Disk deletion is working.
- E2E tests are working.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: disks
type: chore
summary: Fix const for PVCLost phase, sync Go sources and CRDs (VD, CVI).
```

### DIFF
--- a/crds/clustervirtualimages.yaml
+++ b/crds/clustervirtualimages.yaml
@@ -408,8 +408,9 @@ spec:
                   description: Deprecated. Use `imageUploadURLs` instead.
                   type: string
                 usedInNamespaces:
-                  description: |
-                    Displays the list of namespaces where the image is currently used.
+                  description:
+                    Displays the list of namespaces where the image is currently
+                    used.
                   items:
                     type: string
                   type: array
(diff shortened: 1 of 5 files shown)


crds/clustervirtualimages.yaml
```diff
@@ -408,8 +408,6 @@ spec:
                   description: Deprecated. Use `imageUploadURLs` instead.
                   type: string
                 usedInNamespaces:
-                  description: |
```
Isteb4k, line 411: Why did you decide to delete the description for this field?

diafour: Accidental commit after not-synced generation.

danilrwx (approved):

Isteb4k (approved):

PR #1207, opened 2025-06-30
fix(module): fix validating for field https.mode

LopatinDmitr:
## Description

Fix validation logic related to the `https.mode` field in the virtualization module settings, ensuring that the UPLOADER_INGRESS_TLS_SECRET is only required when the HTTPS mode is 'CertManager' or 'CustomCertificate'.

If the `https.mode` field in the module settings is set to `OnlyInURI`, we get an error:
```
  1. ModuleRun:main:virtualization:KubeConfig-Changed-Modules:failures 1:run helm install: execution error at (virtualization/templates/virtualization-controller/deployment.yaml:96:16): https.mode must be CustomCertificate or CertManager
```

All work as expected

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: module
type: fix
summary: Fix validation logic related to the https.mode field in the virtualization module settings, ensuring that the UPLOADER_INGRESS_TLS_SECRET is only required when the HTTPS mode is 'CertManager' or 'CustomCertificate'.
```

### DIFF
--- a/images/virtualization-artifact/pkg/config/load_dvcr_settings.go
+++ b/images/virtualization-artifact/pkg/config/load_dvcr_settings.go
@@ -69,9 +69,6 @@ func LoadDVCRSettingsFromEnvs(controllerNamespace string) (*dvcr.Settings, error
 	if dvcrSettings.UploaderIngressSettings.Host == "" {
 		return nil, fmt.Errorf("environment variable %q undefined, specify DVCR settings", UploaderIngressHostVar)
 	}
-	if dvcrSettings.UploaderIngressSettings.TLSSecret == "" {
-		return nil, fmt.Errorf("environment variable %q undefined, specify DVCR settings", UploaderIngressTLSSecretVar)
-	}
 	if dvcrSettings.UploaderIngressSettings.Class == "" {
 		return nil, fmt.Errorf("environment variable %q undefined, specify DVCR settings", UploaderIngressClassVar)
 	}
--- a/images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
+++ b/images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
@@ -84,7 +84,6 @@ func (i *Ingress) makeSpec() *netv1.Ingress {
 			Namespace: i.Settings.Namespace,
 			Annotations: map[string]string{
 				annotations.AnnUploadURL:                              uploadURL,
-				"nginx.ingress.kubernetes.io/ssl-redirect":            "true",
 				"nginx.ingress.kubernetes.io/proxy-body-size":         "0",
 				"nginx.ingress.kubernetes.io/proxy-request-buffering": "off",
 				"nginx.ingress.kubernetes.io/proxy-buffering":         "off",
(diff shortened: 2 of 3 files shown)


images/virtualization-artifact/pkg/config/load_dvcr_settings.go
```diff
@@ -69,9 +69,6 @@ func LoadDVCRSettingsFromEnvs(controllerNamespace string) (*dvcr.Settings, error
 	if dvcrSettings.UploaderIngressSettings.Host == "" {
 		return nil, fmt.Errorf("environment variable %q undefined, specify DVCR settings", UploaderIngressHostVar)
 	}
-	if dvcrSettings.UploaderIngressSettings.TLSSecret == "" {
```
Isteb4k, line 72: For `Disabled` and `OnlyInURI`, the transmission of bytes is done via HTTP, without TLS.
If we use https.mode for all our ingresses, then it seems logical to create an ingress without TLS for the uploader as well.

Isteb4k: In your case, in the PR, TLS will be requested but without a certificate - I highly doubt that's what we want

deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/16023017652)

templates/virtualization-controller/_helpers.tpl
```diff
@@ -54,8 +54,10 @@
   value: "10m"
 - name: UPLOADER_INGRESS_HOST
   value: {{ include "helm_lib_module_public_domain" (list . "virtualization") }}
+{{- if or (eq (include "helm_lib_module_https_mode" .) "CertManager") (eq (include "helm_lib_module_https_mode" .) "CustomCertificate") }}
```
yaroslavborbat, line 57: use https://github.com/deckhouse/deckhouse/blob/2ee4a924dbec676f4e1c61972f3a5586ca6f1ce3/helm_lib/charts/deckhouse_lib_helm/templates/_module_https.tpl#L112

images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
```diff
@@ -76,7 +83,7 @@ func (i *Ingress) makeSpec() *netv1.Ingress {
 			Name:      i.Settings.Name,
 			Namespace: i.Settings.Namespace,
 			Annotations: map[string]string{
-				annotations.AnnUploadURL:                              fmt.Sprintf("https://%s%s", i.Settings.Host, path),
+				annotations.AnnUploadURL:                              uploadURL,
 				"nginx.ingress.kubernetes.io/ssl-redirect":            "true",
```
yaroslavborbat, line 87: should be set if tls enabled

yaroslavborbat: "nginx.ingress.kubernetes.io/ssl-redirect":            "true",

Isteb4k (approved):

PR #1260, opened 2025-07-17
chore(core): virt-operator placement

universal-itengineer:
## Description

The virt-operator must be placed on the system nodes if they are present in the cluster, otherwise on the master nodes.

The previously used template did not produce the expected result.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: the virt-operator must be placed on the system nodes if they are present in the cluster, otherwise on the master nodes.
```


templates/_helpers.tpl
```diff
@@ -38,3 +38,18 @@ spec:
...
+{{-   if eq $strategy "system" }}
+{{-     if gt (index $context.Values.global.discovery.d8SpecificNodeCountByRole "system" | int) 0 }}
+nodeSelector:
+  node-role.deckhouse.io/system: ""
+{{-     else }}
+nodeSelector:
+  node-role.kubernetes.io/control-plane: ""
+{{-     end }}
+{{-   end }}
+{{- end }}
```
nevermarine, line 55: Maybe the system argument is redundant? We check the number of system nodes via `d8SpecificNodeCountByRole` parameter, and I guess we don't really need the `eq $strategy "system"` check.
```suggestion
{{- /* Returns node selector for workloads depend on strategy. (Returns only system or control-plane node selector) */ -}}
{{- define "virt_helper_node_selector" }}
{{-   $context := index . 0 }} {{- /* Template context with .Values, .Chart, etc */ -}}
{{-   if gt (index $context.Values.global.discovery.d8SpecificNodeCountByRole "system" | int) 0 }}
nodeSelector:
  node-role.deckhouse.io/system: ""
{{-   else }}
nodeSelector:
  node-role.kubernetes.io/control-plane: ""
{{-   end }}
{{- end }}
```

nevermarine (changes requested):

templates/_helpers.tpl
diafour: Without argument the template should be renamed to more specific, e.g. "node_selector_system_or_control_plane"

PR #1292, opened 2025-07-31
refactor: improve gc

yaroslavborbat:
## Description
- GC will run every day at 00:00.
- GC will delete all VMOPs in completed state if their TTL has expired (24 hours).
- GC will delete VMOPs in completed state even if their TTL has not expired, provided there are more than 10 such VMOPs, keeping only the latest 10.

## Checklist
- [X] The code is covered by unit tests.
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: chore
summary: refactor GC.
impact_level: low
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/gc/cron_source.go
+++ b/images/virtualization-artifact/pkg/controller/gc/cron_source.go
@@ -41,7 +41,7 @@ type SourceGCManager interface {
 	ListForDelete(ctx context.Context, now time.Time) ([]client.Object, error)
 }
 
-func NewCronSource(scheduleSpec string, client client.Client, mgr SourceGCManager, log *log.Logger) (*CronSource, error) {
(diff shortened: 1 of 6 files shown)


images/virtualization-artifact/pkg/controller/gc/cron_source.go
```diff
@@ -17,157 +17,92 @@ limitations under the License.
...
+}
 
-		if err := meta.SetList(objs, expiredItems); err != nil {
-			log.Error("failed to set list", logger.SlogErr(err))
-		}
-		return objs
-	}
+type CronSource struct {
+	schedule cron.Schedule
+	client   client.Client
```
diafour, line 61: Is `client` needed? It seems CrontSource do not call the client. Everything `client`-related is hidden into `mgr`, so client is not needed anymore.

PR #1324, opened 2025-08-07
chore(module): rewrite module_config_validator.py hook in Go

diafour:
## Description

- Rewrite last hook, no more Python hooks.
- Add readiness probe to not block the 'main' queue.

We don't want to block a 'main' queue when user enable virtualization module with invalid CIDRs (overlaps with node addresses, podSubnet or serviceSubnet).

'main' queue is not blocked if virtualization module was enabled with invalid configuration.

Снимок экрана 2025-08-12 в 02 38 14

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: module
type: chore
summary: Rewrite module_config_validator.py hook in Go, remove python dependencies from the build.
```


templates/virtualization-audit/_helper.tpl
```diff
@@ -0,0 +1,21 @@
...
+Usage:
+
+    {{- if eq (include "audit.isEnabled" .) "true" }}
+    kind: Deployment
+    ...
+    {{ -end }}
+
+*/}}
+{{- define "audit.isEnabled" -}}
+{{- if (dig "moduleConfig" "audit" "enabled" false .Values.virtualization.internal) -}}
```
yaroslavborbat, line 14: false?

z9r5 (dismissed):

yaroslavborbat (approved):

diafour:
These changes was tested in dev cluster with enabling and disabling module. No changes that affect VM/VI/VD, etc.

PR #1408, opened 2025-08-29
fix(vm): block network changes when SDN is disabled

loktev-d:
## Description

Add validation to prevent modifications to .spec.networks field in VirtualMachine resources when the SDN feature gate is disabled.

Currently, users can add additional network interfaces to VMs even when SDN is disabled. While the reconciliation loop detects this and sets an error condition, the spec change is still accepted and stored. This creates confusion as the VM enters an error state after the change is already applied.

By adding webhook validation, we now reject invalid network configurations at admission time, preventing the spec from being modified in the first place. This provides clearer, immediate feedback to users about why their change is not allowed.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: Block network spec changes when SDN feature gate is disabled
```


deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/17358424679)

yaroslavborbat (approved):

PR #1414, opened 2025-08-31
fix(vm): fix message in `NetworkReady` condition

LopatinDmitr:
## Description

Fix message in NetworkReady condition:
- when module SDN don't set annotation with network status data
- when vm pods not exist

Message "Annotation %s is not found" is not clear to the user.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: fix message in NetworkReady condition
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/network.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/network.go
@@ -153,6 +153,10 @@ func (h *NetworkInterfaceHandler) UpdateNetworkStatus(ctx context.Context, s sta
 func extractNetworkStatusFromPods(pods *corev1.PodList) (string, error) {
 	var errorMessages []string
 
+	if len(pods.Items) == 0 {
+		return "Waiting for the pod to be created.", nil
+	}
+
 	for _, pod := range pods.Items {
 		if pod.Status.Phase == corev1.PodSucceeded {
 			continue
@@ -160,7 +164,7 @@ func extractNetworkStatusFromPods(pods *corev1.PodList) (string, error) {
 
 		networkStatusAnnotation, found := pod.Annotations[annotations.AnnNetworksStatus]
 		if !found {
-			errorMessages = append(errorMessages, fmt.Sprintf("Annotation %s is not found", annotations.AnnNetworksStatus))
+			errorMessages = append(errorMessages, "Cannot determine the status of additional interfaces, waiting for a response from the SDN module")
 			continue
 		}
 


LopatinDmitr:
image

images/virtualization-artifact/pkg/controller/vm/internal/network.go
```diff
@@ -152,6 +152,11 @@ func (h *NetworkInterfaceHandler) UpdateNetworkStatus(ctx context.Context, s sta
 
 func extractNetworkStatusFromPods(pods *corev1.PodList) (string, error) {
 	var errorMessages []string
+	templateErrorMessage := "Network status data not found - please check SDN module operation and contact administrator if needed."
```
Isteb4k, line 155: Move into a const outside the function

images/virtualization-artifact/pkg/controller/vm/internal/network.go
```diff
@@ -152,6 +152,11 @@ func (h *NetworkInterfaceHandler) UpdateNetworkStatus(ctx context.Context, s sta
...
+
+	if len(pods.Items) == 0 {
+		return templateErrorMessage, nil
```
Isteb4k, line 158: How is the absence of pods related to the absence of the SDN module?

images/virtualization-artifact/pkg/controller/vm/internal/network.go
```diff
@@ -160,7 +165,7 @@ func extractNetworkStatusFromPods(pods *corev1.PodList) (string, error) {
 
 		networkStatusAnnotation, found := pod.Annotations[annotations.AnnNetworksStatus]
 		if !found {
-			errorMessages = append(errorMessages, fmt.Sprintf("Annotation %s is not found", annotations.AnnNetworksStatus))
+			errorMessages = append(errorMessages, templateErrorMessage)
```
Isteb4k, line 168: Maybe a better "Cannot determine the status of additional interfaces, waiting for a response from the SDN module"?

Isteb4k, line 168: Maybe it’s better to explicitly check whether the SDN module is enabled in the cluster — through the `Module` resource

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vm/internal/network.go
LopatinDmitr: Fixed.

images/virtualization-artifact/pkg/controller/vm/internal/network.go
```diff
@@ -153,14 +153,18 @@ func (h *NetworkInterfaceHandler) UpdateNetworkStatus(ctx context.Context, s sta
 func extractNetworkStatusFromPods(pods *corev1.PodList) (string, error) {
 	var errorMessages []string
 
+	if len(pods.Items) == 0 {
+		return "Waiting for pod to be created", nil
```
Isteb4k, line 157: If this message is for a condition, then it should end with a period.

Isteb4k, line 157: for **the** pod

Isteb4k (changes requested):

images/virtualization-artifact/pkg/controller/vm/internal/network.go
LopatinDmitr: Fixed.

PR #1417, opened 2025-09-01
fix(vd): respect user-specified storage class when restoring from snapshot

loktev-d:
## Description

Fix the VirtualDisk controller to respect user-specified storageClassName when creating disks from VirtualDiskSnapshot. Previously, the controller was always using the storage class from the original disk (stored in VolumeSnapshot annotations), completely ignoring the user-specified value in `spec.persistentVolumeClaim.storageClassName`. Also add error message when user tries to perform cross-provider restore.

 When users tried to restore a VirtualDisk from a snapshot with a different storage class, the parameter `spec.persistentVolumeClaim.storageClassName` was silently ignored. The restored disk would always use the same storage class as the original disk.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vd
type: fix
summary: respect user-specified storage class when restoring from snapshot
```


images/virtualization-artifact/pkg/controller/vd/internal/source/step/create_pvc_from_vdsnapshot_step.go
```diff
@@ -219,3 +241,44 @@ func (s CreatePVCFromVDSnapshotStep) buildPVC(vd *virtv2.VirtualDisk, vs *vsv1.V
...
+	}
+
+	targetSCName := *vd.Spec.PersistentVolumeClaim.StorageClass
+
+	originalSCName := vs.Annotations[annotations.AnnStorageClassName]
+	if originalSCName == "" {
+		originalSCName = vs.Annotations[annotations.AnnStorageClassNameDeprecated]
+	}
+
+	if targetSCName == originalSCName {
```
hardcoretime, line 257: Can this statement be a problem if the target storage class has a different provisioner than the original?
For example, a volume snapshot is created with the original storage class, but after snapshotting, the original storage class is replaced by another storage class with the same name but a different provisioner.

deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/18096632567)

deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/18121846153)

hardcoretime (approved):

hardcoretime (approved):

PR #1490, opened 2025-09-22
test(vmrestore): disable checks until issues are fixed

hardcoretime:
## Description

- Skip test if forced restoration fails with the virtual machine in the pending phase.
- Skip annotation and label checks.
- Skip MAC address checks.

These bugs are currently known, and nightly end-to-end tests should not fail for the related cases.

Nightly end-to-end tests should not fail due to known errors.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: api
type: chore
summary: "Nightly end-to-end tests should not fail due to known errors."
impact_level: low
```

### DIFF
--- a/tests/e2e/vm_restore_force_test.go
+++ b/tests/e2e/vm_restore_force_test.go
@@ -49,6 +49,7 @@ var _ = Describe("VirtualMachineRestoreForce", SIGRestoration(), ginkgoutil.Comm
 		testCaseLabel       = map[string]string{"testcase": "vm-restore-force"}
 		additionalDiskLabel = map[string]string{"additionalDisk": "vm-restore-force"}
 		originalVMNetworks  map[string][]virtv2.NetworksStatus
+		criticalError       string
 	)
 
 	BeforeAll(func() {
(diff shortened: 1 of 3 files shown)


tests/e2e/vm_restore_force_test.go
```diff
@@ -232,14 +233,48 @@ var _ = Describe("VirtualMachineRestoreForce", SIGRestoration(), ginkgoutil.Comm
...
+						}
+
+						readyCondition, err := GetCondition(vmrestorecondition.VirtualMachineRestoreReady.String(), vmRestoreObj)
+						if err != nil {
+							return err
+						}
+
+						msg := "A virtual machine cannot be restored from the pending phase with `Forced` mode; you can delete the virtual machine and restore it with `Safe` mode."
+						if vmRestoreObj.Status.Phase == virtv2.VirtualMachineRestorePhaseFailed && readyCondition.Message == msg {
+							Skip("A bug has occurred with a virtual machine in the \"Pending\" phase.")
```
Isteb4k, line 258: If only a specific By is skipped, it looks as though in the case of the 'vm Pending' error, the test will fail at the 'Checking the result of restoration' step, since the disks will not start restoring

Isteb4k (changes requested):

Isteb4k (dismissed):

PR #1534, opened 2025-10-06
feat(core): promote vd/vi import error if url is broken

eofff:
## Description

Promote vd/vi import error if url is broken.

Right now, if the HTTP source returns an error, the VI/VD will hang in the “Provisioning” status, and the only way to tell that something went wrong will be by checking the importer pod logs.

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: core
type: feature
summary: Promote vd/vi import error if url is broken.
```


deckhouse-BOaTswain:
Workflow has started.
         Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/18287385509)

PR #2083, opened 2026-03-06
 fix(dra): handle parent watcher channel close in wrapWatcher

yaroslavborbat:
## Description

Fix `wrapWatcher` in the virtualization-dra plugin so that it correctly reacts when the parent watcher closes its `ResultChan()`. Previously the wrapper did not check the closed state of the parent channel, so events stopped being delivered and the cache stopped updating.

Changes:
- Check `ok` when reading from the parent's `ResultChan()`; when the channel is closed (`ok == false`), exit the receive loop and close the wrapper's channels.
- Add a `done` channel that is closed when the receive goroutine exits; close both `result` and `done` in `receive()` via defer.
- In `Stop()`, wait for `<-w.done` so that the receive goroutine fully exits before `Stop()` returns.
- When sending to `w.result`, use a select with `ctx.Done()` to avoid blocking if the consumer has stopped reading.
- Change `newWrapWatcher` return type to `*wrapWatcher` so callers can use the concrete type (e.g. for proper `Stop()` semantics).

When the parent watcher closes its result channel (e.g. on reconnect, apiserver disconnect, or explicit stop), the wrapper was not detecting it. The code used `event := <-resultChan` without the second `ok` value, so after the parent closed the channel the wrapper could either receive only zero-value events or block indefinitely. As a result:

- The wrapper's `result` channel was never closed.
- Consumers (including cache/listwatch logic) did not see the stream as finished and did not refresh or resync.
- The cache could become stale and not receive new events after the parent watcher was replaced or closed.

The fix ensures that when the parent closes its channel, the wrapper closes its own channels, so consumers are notified and can react (e.g. relist or re-establish the watch), keeping the cache in sync.

- When the parent watcher closes `ResultChan()`, the wrapper's `ResultChan()` is closed as well.
- Consumers detect the closed channel and can relist or re-watch as needed.
- The cache is updated correctly after reconnects or parent watcher restarts.
- `Stop()` waits for the receive goroutine to finish, so there are no goroutine leaks and cleanup order is predictable.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-dra/internal/plugin/wrapresourceslice/watcher.go
+++ b/images/virtualization-dra/internal/plugin/wrapresourceslice/watcher.go
@@ -22,7 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
-func newWrapWatcher(ctx context.Context, w watch.Interface, match func(event watch.Event) bool) watch.Interface {
+func newWrapWatcher(ctx context.Context, w watch.Interface, match func(event watch.Event) bool) *wrapWatcher {
 	ctx, cancel := context.WithCancel(ctx)
 
 	watcher := &wrapWatcher{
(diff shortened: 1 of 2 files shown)


yaroslavborbat:
https://github.com/deckhouse/virtualization/pull/2085

PR #2182, opened 2026-04-02
feat(vmop): add migration progress status and reasons

danilrwx:
## Description
Implement VMOP migration progress reporting and make sync-phase progress more informative and stable.

Changes in this PR:
- add `status.progress` support for migration `VirtualMachineOperation` objects;
- add detailed `Completed` reasons for migration lifecycle stages;
- centralize sync progress range in `migrationprogress.SyncRangeMin/SyncRangeMax`;
- add a dedicated migration progress package with mapper/strategy logic;
- consume KubeVirt migration transfer details from `status.migrationState.transferStatus` when they are available;
- keep degraded time/phase-based progress as a fallback when transfer details are unavailable;
- reduce repeated artificial stall bumps in degraded mode.

Migration VMOP progress was previously coarse and bursty. In practice users could observe sequences like `1 -> 2 -> 3 -> 16 -> 41 -> 42..51 -> 100`, where the visible progress was driven by reconcile cadence and fallback estimation rather than by the actual migration state.

This PR improves the user-facing progress model by:
- exposing progress directly in VMOP status;
- providing stage-specific reasons such as disks preparation, target scheduling, target preparation, syncing, source suspended, and target resumed;
- improving sync-phase progress calculation so it can consume KubeVirt migration transfer counters and runtime signals from `transferStatus` when they are available;
- keeping the old degraded estimator as a compatibility fallback;
- preventing the previous near `+1 per reconcile` behavior from the stall bump logic.

As a result, migration progress becomes more actionable for users and easier to debug from VMOP conditions.

This PR depends on the matching `3p-kubevirt` API change that moves migration transfer-related fields under `VirtualMachineInstanceMigrationState.transferStatus`.

1. Start a VM migration through `VirtualMachineOperation`.
2. Observe `status.progress` on the VMOP object.
3. During early lifecycle phases, progress should move through the coarse stage milestones.
4. During sync, progress should stay within the configured sync range and advance more naturally.
5. When KubeVirt exposes `status.migrationState.transferStatus`, sync progress should use those runtime values.
6. `Completed` condition reasons should reflect the current migration stage.
7. In environments where migration transfer details are not available yet, degraded fallback should still work without repeated artificial increments on every reconcile.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: vm
type: feature
summary: "Added the `progress` field to the status of `VirtualMachineOperation` resources with the `Evict` and `Migrate` types to show operation progress. The corresponding `PROGRESS` column is displayed when running `d8 k get vmop`"
```


images/virtualization-artifact/pkg/monitoring/metrics/vmop/data_metric.go
```diff
@@ -36,6 +36,20 @@ type dataMetric struct {
...
+}
+
+var failedTerminalReasons = map[string]struct{}{
+	string(vmopcondition.ReasonOperationFailed):     {},
+	string(vmopcondition.ReasonFailed):              {},
+	string(vmopcondition.ReasonAborted):             {},
+	string(vmopcondition.ReasonNotConverging):       {},
+	string(vmopcondition.ReasonTargetUnschedulable): {},
+	string(vmopcondition.ReasonTargetDiskError):     {},
+}
```
yaroslavborbat, line 51: .String() ?

yaroslavborbat (dismissed):

yaroslavborbat (approved):

PR #2216, opened 2026-04-14
fix(module): add idempotency for usb cleanup in virtualization-dra

danilrwx:
## Description

Add idempotency to USB device cleanup in the virtualization-dra component.

Changes:
- **`store.go`**: Add early return if claim is already unprepared. Use `hasCount` check to handle missing or zero counts gracefully. Replace `if/else` with `switch` for clearer count handling (0, 1, >1).
- **`usbgateway.go`**: Make `Detach` idempotent — log and skip if device is already detached instead of failing. Always clean up the attach record after detach attempt.
- **`attach_record.go`**: Add `RemoveEntryByDeviceName` method to remove a specific device entry from the attach record. Extract `storeLocked` helper to avoid code duplication.

When `UnprepareResources` is called multiple times for the same claim (e.g., during retries or race conditions), the cleanup logic could fail or produce incorrect results:
- Detaching an already-detached device caused errors.
- The attach record was not cleaned up if the device was already detached.
- The `usbipAllocatedDevicesCount` map could be accessed for non-existent keys, leading to incorrect behavior.

After this fix, repeated cleanup calls are safe and produce consistent results.

1. Call `UnprepareResources` for a claim that has USB devices attached via USBGateway.
2. Devices are detached and attach records are removed.
3. Call `UnprepareResources` again for the same claim.
4. No errors occur — the operation is a no-op.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: module
type: fix
summary: Add idempotency for USB device cleanup in virtualization-dra to prevent errors on repeated unprepare calls.
impact_level: low
```

### DIFF
--- a/images/virtualization-dra/internal/usb-gateway/attach_record.go
+++ b/images/virtualization-dra/internal/usb-gateway/attach_record.go
@@ -118,19 +118,20 @@ func (r *attachRecordManager) Refresh() error {
 		return err
 	}
 
-	// keep only real entries
-	var newEntries []AttachEntry
+	newEntries := make([]AttachEntry, 0, len(record.Entries))
 	for _, e := range record.Entries {
 		if _, ok := ports[e.Rhport]; ok {
 			newEntries = append(newEntries, e)
 		}
 	}
 
-	record.Entries = newEntries
+	newRecord := attachRecord{Entries: newEntries}
+	if slices.Equal(record.Entries, newRecord.Entries) {
+		r.record = newRecord
+		return nil
+	}
 
-	r.record = record
-
-	return nil
+	return r.storeLocked(newRecord)
 }
 
 func (r *attachRecordManager) GetEntries() []AttachEntry {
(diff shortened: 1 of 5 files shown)


images/virtualization-dra/internal/usb-gateway/attach_record.go
```diff
@@ -163,8 +164,10 @@ func (r *attachRecordManager) AddEntry(e AttachEntry) error {
 	newEntries := slices.Clone(r.record.Entries)
 	newEntries = append(newEntries, e)
 
-	record := attachRecord{Entries: newEntries}
+	return r.storeLocked(attachRecord{Entries: newEntries})
+}
 
+func (r *attachRecordManager) storeLocked(record attachRecord) error {
```
yaroslavborbat: why locked?

images/virtualization-dra/internal/usb-gateway/attach_record.go
```diff
@@ -118,19 +118,20 @@ func (r *attachRecordManager) Refresh() error {
 		return err
 	}
 
-	// keep only real entries
```
yaroslavborbat, line 121: keep comment

images/virtualization-dra/internal/usb-gateway/usbgateway.go
```diff
@@ -101,7 +101,9 @@ func (c *USBGatewayController) Detach(deviceName string) error {
 	}
 
 	entry := c.findEntry(deviceName)
-	if entry != nil {
+	if entry == nil {
+		log.Info("Device is already detached")
```
yaroslavborbat, line 105: Can we return here? If the record doesn’t exist, doesn’t that guarantee that we have already performed detach and unexport?

yaroslavborbat (changes requested):

yaroslavborbat (approved):

PR #2221, opened 2026-04-15
feat(nodeusbdevice): mirror usb attached condition

danilrwx:
## Description
Add the `Attached` condition to `NodeUSBDevice` and mirror it from `USBDevice.status.conditions[type=Attached]`.

The change also adds an `Attached` printer column for `NodeUSBDevice`, updates the controller to reconcile this condition, and adds a watcher so `NodeUSBDevice` is refreshed when the corresponding `USBDevice` changes.

`NodeUSBDevice` currently exposes only `Ready` and `Assigned` conditions, so it is not possible to tell from this resource whether the device is currently attached to a virtual machine.

The actual attachment state already exists in `USBDevice`, but it is inconvenient for cluster-level device visibility because `NodeUSBDevice` is the cluster-scoped source operators look at first.

This change makes `NodeUSBDevice` reflect the existing attachment state without introducing a second source of truth. The controller simply mirrors `USBDevice.Attached` status, reason, and message into `NodeUSBDevice.Attached`.

1. Assign a `NodeUSBDevice` to a namespace.
2. Before the device is attached to a VM, `kubectl get nodeusbdevices` shows `Attached=False`.
3. Attach the corresponding `USBDevice` to a `VirtualMachine`.
4. `NodeUSBDevice.status.conditions[type=Attached]` becomes `True` and the `Attached` printer column is updated.
5. When the device is detached or temporarily unavailable for migration, `NodeUSBDevice` mirrors the corresponding `USBDevice.Attached` state and reason.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher.go
+++ b/images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher.go
@@ -46,12 +46,6 @@ func (w *USBDeviceWatcher) Watch(mgr manager.Manager, ctr controller.Controller)
 				return requestsByUSBDevice(usbDevice)
 			}),
 			predicate.TypedFuncs[*v1alpha2.USBDevice]{
-				CreateFunc: func(e event.TypedCreateEvent[*v1alpha2.USBDevice]) bool {
-					return e.Object != nil
-				},
-				DeleteFunc: func(e event.TypedDeleteEvent[*v1alpha2.USBDevice]) bool {
-					return e.Object != nil
-				},
 				UpdateFunc: func(e event.TypedUpdateEvent[*v1alpha2.USBDevice]) bool {
 					return shouldProcessUSBDeviceUpdate(e.ObjectOld, e.ObjectNew)
 				},
--- a/images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher.go
+++ b/images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher.go
@@ -1,0 +1,91 @@
+/*
+Copyright 2026 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package watcher
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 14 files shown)


images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher.go
```diff
@@ -0,0 +1,91 @@
...
+			handler.TypedEnqueueRequestsFromMapFunc(func(_ context.Context, usbDevice *v1alpha2.USBDevice) []reconcile.Request {
+				return requestsByUSBDevice(usbDevice)
+			}),
+			predicate.TypedFuncs[*v1alpha2.USBDevice]{
+				CreateFunc: func(e event.TypedCreateEvent[*v1alpha2.USBDevice]) bool {
+					return e.Object != nil
+				},
+				DeleteFunc: func(e event.TypedDeleteEvent[*v1alpha2.USBDevice]) bool {
+					return e.Object != nil
+				},
```
yaroslavborbat: Object can be nil? Delete this predicates

images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher_test.go
```diff
@@ -0,0 +1,97 @@
...
+	requests := requestsByUSBDevice(usbDevice)
+	if len(requests) != 1 {
+		t.Fatalf("expected 1 request, got %d", len(requests))
+	}
+	if requests[0].Name != "node-usb-device-1" {
+		t.Fatalf("unexpected request name %q", requests[0].Name)
+	}
+	if requests[0].Namespace != "" {
+		t.Fatalf("expected cluster-scoped request, got namespace %q", requests[0].Namespace)
+	}
```
yaroslavborbat: assert

images/virtualization-artifact/pkg/controller/nodeusbdevice/internal/watcher/usbdevice_watcher_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestRequestsByUSBDeviceFallsBackToName(t *testing.T) {
+	usbDevice := &v1alpha2.USBDevice{ObjectMeta: metav1.ObjectMeta{Name: "usb-device-1", Namespace: "test-ns"}}
+
+	requests := requestsByUSBDevice(usbDevice)
+	if len(requests) != 1 {
+		t.Fatalf("expected 1 request, got %d", len(requests))
+	}
+	if requests[0].Name != "usb-device-1" {
+		t.Fatalf("unexpected fallback request name %q", requests[0].Name)
+	}
```
yaroslavborbat: assert

yaroslavborbat: and please check other tests

test/e2e/vm/usb.go
```diff
@@ -238,6 +266,14 @@ func (t *VMUSBTest) mountUSB() {
 	Expect(err).NotTo(HaveOccurred())
 }
 
+func nodeUSBAttachedCondition(nodeUSBDevice *v1alpha2.NodeUSBDevice) *metav1.Condition {
+	if nodeUSBDevice == nil {
+		return nil
+	}
+
+	return meta.FindStatusCondition(nodeUSBDevice.Status.Conditions, "Attached")
```
yaroslavborbat: use const

yaroslavborbat (changes requested):

yaroslavborbat (approved):

PR #2243, opened 2026-04-20
fix(module): avoid queue blocking on invalid module config in CSE

LopatinDmitr:
## Description
Make virtualization module config validation tolerate empty or incomplete `spec.settings` and ensure hooks that patch values only reconcile when a valid copied module config exists.

This change keeps `validate-module-config` as the single source of truth for config validity:
- valid settings are copied into `virtualization.internal.moduleConfig`
- invalid settings are not copied
- the previous valid copied config is preserved
- validation errors are reported through `virtualization.internal.moduleConfigValidation`

Hooks and templates that depend on module config now use the copied internal config contract consistently.

During installation the `ModuleConfig/virtualization` object may appear before the validating webhook is ready, so an empty or incomplete `spec.settings` can temporarily exist in the cluster.

Previously this could lead to two failure modes:
1. `validate-module-config` could panic while accessing missing required settings.
2. Hooks that patch values could continue reconciling from an invalid state, return invalid values patches, and block the module queue when Deckhouse validated them against OpenAPI.

With this change, config validation errors are reported cleanly through readiness, invalid settings never overwrite the copied internal config, and hooks that patch values only run when a valid copied config is available.

1. Create or keep `ModuleConfig/virtualization` with empty or incomplete `spec.settings`.
2. Verify `validate-module-config` does not panic and stores a validation error.
3. Verify invalid module settings are not copied into `virtualization.internal.moduleConfig`.
4. Verify the previous valid copied config is preserved if a new invalid config is applied.
5. Verify hooks that patch values do not reconcile until a valid copied config exists, and continue to use the previous valid copied config if it already exists.
6. Verify the module reports a clear readiness error for the invalid config.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: module
type: fix
summary: Make virtualization hooks use only valid copied module config and avoid queue blocking on invalid module settings.
impact_level: low
```

### DIFF
--- a/images/hooks/pkg/hooks/validate-module-config/hook.go
+++ b/images/hooks/pkg/hooks/validate-module-config/hook.go
@@ -101,17 +101,22 @@ func Reconcile(_ context.Context, input *pkg.HookInput) error {
 		input.Values.Set(settings.InternalValuesConfigValidationPath, map[string]string{
 			"error": fmt.Sprintf("ModuleConfig/virtualization is invalid: %v", err),
 		})
-	} else {
-		// Module is valid, remove moduleConfigValidation object to indicate valid state for the readiness probe.
-		input.Values.Remove(settings.InternalValuesConfigValidationPath)
-		// Copy valid settings from config values to apply them in helm templates.
-		copyModuleConfigSettingsIntoInternalValues(input)
+		return nil
 	}
+
+	// Module is valid, remove moduleConfigValidation object to indicate valid state for the readiness probe.
+	input.Values.Remove(settings.InternalValuesConfigValidationPath)
+	// Copy valid settings from config values to apply them in helm templates.
+	copyModuleConfigSettingsIntoInternalValues(input)
 
 	return nil
 }
 
 func validateModuleConfigSettings(mc *mcapi.ModuleConfig, nodes []corev1.Node, podSubnet, serviceSubnet netip.Prefix) error {
+	if err := validateRequiredModuleConfigSettings(mc); err != nil {
+		return err
+	}
+
 	CIDRs, err := moduleconfig.ParseCIDRs(mc.Spec.Settings)
 	if err != nil {
 		return err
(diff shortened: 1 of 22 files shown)


templates/_helpers.tpl
```diff
@@ -62,7 +62,7 @@ nodeSelector:
 {{- end }}
 
 {{- define "hasValidModuleConfig" -}}
-{{- if (hasKey .Values.virtualization.internal "moduleConfig" ) -}}
+{{- if and (hasKey .Values.virtualization.internal "moduleConfig") (not (hasKey .Values.virtualization.internal "moduleConfigValidation")) -}}
```
diafour: It may have a previous valid moduleConfig and an error for new invalid moduleConfig in moduleConfigValidation, so hasKey "moduleConfig" is enough.
Invalid module settings should not be copied into moduleConfig - that is the contract for the validate-module-config hook.

images/hooks/pkg/hooks/validate-module-config/hook.go
```diff
@@ -101,17 +101,23 @@ func Reconcile(_ context.Context, input *pkg.HookInput) error {
...
-		copyModuleConfigSettingsIntoInternalValues(input)
+		input.Values.Remove(settings.InternalValuesConfigCopyPath)
+		return nil
 	}
 
+	// Module is valid, remove moduleConfigValidation object to indicate valid state for the readiness probe.
+	input.Values.Remove(settings.InternalValuesConfigValidationPath)
+	// Copy valid settings from config values to apply them in helm templates.
+	copyModuleConfigSettingsIntoInternalValues(input)
+
```
diafour: This breaks the contract: "moduleConfig" may have previous valid config, we should not remove it from values.

LopatinDmitr: Fixed

PR #2253, opened 2026-04-21
fix(cli): refresh token lost on reconnect

danilrwx:
## Description
Refresh CLI client config before reconnecting long-lived VM access sessions.

This PR updates `d8 v vnc` and `d8 v console` so they recreate the Kubernetes client from the current context before each reconnect attempt instead of reusing the client created at command start.

Also adds unit tests covering reconnect behavior for both commands.

When `d8 v vnc` or `d8 v console` stays alive for a long time, the process keeps the old in-memory OIDC config. If the ID token expires and another client (`kubectl`, another `d8`, Lens, k9s, IDE) refreshes the token first, Dex rotates the refresh token and the old CLI process can no longer use its stale refresh token on reconnect.

As a result, after VNC/console reconnect scenarios the command may fail with an error like:

```text
failed to refresh token: oauth2: "invalid_request" "Refresh token is invalid or has already been claimed by another client."
```

This change makes reconnect logic reread kubeconfig/client config before each new connection attempt, so the command uses the latest valid tokens.

Reproduction steps for the original problem:
1. Use an OIDC kubeconfig generated by Deckhouse `user-authn` / Dex.
2. Reduce `user-authn.settings.idTokenTTL`, for example to `1m`.
3. Start a long-lived session:
   - `d8 v vnc <vm> -n <ns>`, or
   - `d8 v console <vm> -n <ns>`.
4. Wait until the ID token expires.
5. In another terminal, run `kubectl get ns` to refresh tokens and rotate the refresh token.
6. Trigger reconnect of the original session:
   - reboot the guest OS,
   - or restart the VM,
   - or cause a temporary connection loss.

Expected result after this PR:
- `d8 v vnc` reconnects successfully.
- `d8 v console` reconnects successfully.
- The command does not fail with the stale refresh token error.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: core
type: fix
summary: "Reconnect logic in d8 v vnc and d8 v console now refreshes client config before reconnect, avoiding stale OIDC refresh token failures."
impact_level: low
```


src/cli/internal/cmd/console/console_test.go
```diff
@@ -0,0 +1,152 @@
...
+	err = (&Console{timeout: time.Second}).Run(cmd, []string{"test-vm"})
+	if err != nil {
+		t.Fatalf("Run returned error: %v", err)
+	}
+	if connectCalls != 2 {
+		t.Fatalf("expected 2 connect attempts, got %d", connectCalls)
+	}
+	if clientCalls != 2 {
+		t.Fatalf("expected client to be refreshed before each reconnect, got %d calls", clientCalls)
+	}
```
yaroslavborbat: assert

src/cli/internal/cmd/vnc/vnc_test.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+func (f *fakeKubeclient) NodeUSBDevices() virtualizationv1alpha2.NodeUSBDeviceInterface {
+	return nil
+}
+
+func (f *fakeKubeclient) USBDevices(namespace string) virtualizationv1alpha2.USBDeviceInterface {
+	return nil
+}
```
yaroslavborbat: Don’t we have a generated fake client?

src/cli/internal/cmd/vnc/vnc_test.go
```diff
@@ -0,0 +1,147 @@
...
+		connectCalls++
+		if ln == nil {
+			t.Fatal("listener must not be nil")
+		}
+		if namespace != "default" {
+			t.Fatalf("unexpected namespace: %s", namespace)
+		}
+		if vmName != "test-vm" {
+			t.Fatalf("unexpected vm name: %s", vmName)
+		}
```
yaroslavborbat: assert

src/cli/internal/cmd/vnc/vnc_test.go
```diff
@@ -0,0 +1,147 @@
...
+	err := (&VNC{}).Run(cmd, []string{"test-vm"})
+	if err != nil {
+		t.Fatalf("Run returned error: %v", err)
+	}
+	if connectCalls != 2 {
+		t.Fatalf("expected 2 connect attempts, got %d", connectCalls)
+	}
+	if clientCalls != 2 {
+		t.Fatalf("expected client to be refreshed before each reconnect, got %d calls", clientCalls)
+	}
```
yaroslavborbat: assert

src/cli/internal/cmd/console/console_test.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+func (f *fakeKubeclient) NodeUSBDevices() virtualizationv1alpha2.NodeUSBDeviceInterface {
+	return nil
+}
+
+func (f *fakeKubeclient) USBDevices(namespace string) virtualizationv1alpha2.USBDeviceInterface {
+	return nil
+}
```
yaroslavborbat: Don’t we have a generated fake client?

src/cli/internal/cmd/console/console_test.go
```diff
@@ -0,0 +1,152 @@
...
+	defer func() {
+		os.Stdin = oldStdin
+		clientAndNamespaceFromContext = oldClientAndNamespaceFromContext
+		connectFunc = oldConnectFunc
+	}()
+
+	stdinReader, stdinWriter, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("create stdin pipe: %v", err)
+	}
```
yaroslavborbat: assert

src/cli/internal/cmd/console/console_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	var connectCalls int
+	connectFunc = func(_ context.Context, name, namespace string, _ kubeclient.Client, _ time.Duration, _ <-chan []byte, _ <-chan struct{}) error {
+		connectCalls++
+		if namespace != "default" {
+			t.Fatalf("unexpected namespace: %s", namespace)
+		}
+		if name != "test-vm" {
+			t.Fatalf("unexpected vm name: %s", name)
+		}
```
yaroslavborbat: assert

yaroslavborbat (changes requested):

yaroslavborbat (approved):

PR #2268, opened 2026-04-23
test(vm): port VM placement scenarios to new framework

danilrwx:
## Description
Port the remaining `VirtualMachineAffinityAndToleration` legacy scenarios to the new e2e framework under `test/e2e/vm`.

The PR now covers all placement-related checks from the former legacy suite:
- affinity / anti-affinity placement checks via `status.nodeName`
- migration after affinity updates
- `spec.nodeSelector` behavior
- `spec.affinity.nodeAffinity` behavior

The new tests use the modern framework helpers and object builders, add cluster prerequisite checks for required node topology, and keep migration verification based on fresh `MigrationState` data and resulting `status.nodeName`.

The PR also includes a follow-up fix for the migrated placement scenarios: the selected target node is now preserved across `By(...)` steps, so the final assertions validate the actual migration destination instead of an empty outer-scope value.

The obsolete legacy file was removed after the remaining scenarios were ported.

The old legacy affinity/toleration suite had already been partially migrated, but two placement scenarios still existed only in `test/e2e/legacy/affinity_toleration.go`.

Keeping the coverage split between legacy and the new framework made the suite harder to maintain and left outdated test infrastructure in place. This PR completes the migration, consolidates the scenarios in the new framework, and removes the leftover legacy implementation.

During validation, a scoping bug was also found in the new placement checks: `targetNode` was shadowed inside a step, which caused the outer variable to remain empty and made the final assertion fail despite successful migration. The fix makes the assertions stable and aligned with the actual migrated node.

1. Run the e2e suite that includes `test/e2e/vm/affinity_toleration.go`.
2. Verify that VM placement checks are performed through `status.nodeName` in the new framework.
3. Verify that changing `spec.nodeSelector` to the current node does not trigger migration.
4. Verify that changing `spec.nodeSelector` to another ready worker node triggers migration and updates `status.nodeName`.
5. Verify that changing `spec.affinity.nodeAffinity` to the current node does not trigger migration.
6. Verify that changing `spec.affinity.nodeAffinity` to another ready worker node triggers migration and updates `status.nodeName`.
7. Verify that final assertions use the real target node selected during migration.
8. Verify that the legacy test file is no longer needed.

Additionally, a compile-level validation was executed:
- `cd test/e2e && go test ./... -run '^$'`

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: test
type: chore
summary: "Fix VM placement e2e coverage by migrating remaining legacy scenarios and preserving selected target nodes in final assertions."
impact_level: low
```


test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+	masterLabelKey              = "node.deckhouse.io/group"
+	kvmEnabledLabelKey          = "virtualization.deckhouse.io/kvm-enabled"
+	vmLabelKey                  = "vm"
+	placementNoMigrationWait    = 20 * time.Second
+	placementNoMigrationPolling = time.Second
+)
+
+var _ = Describe("VirtualMachineAffinityAndToleration", func() {
+	var (
+		f   = framework.NewFramework("vm-affinity-toleration")
```
hardcoretime: Move this to `BeforeAll` to prevent initialization during dry-run mode, which will be used in the pre-checks.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+		f.Before()
+		ctx = context.Background()
+	})
+
+	It("checks placement via status.nodeName and migrations after affinity changes", func() {
+		By("Checking test prerequisites", func() {
+			readyNodes, err := listReadyNodes(ctx, f, map[string]string{kvmEnabledLabelKey: "true"})
+			Expect(err).NotTo(HaveOccurred())
+			if len(readyNodes) < 2 {
+				Skip("at least two ready KVM-enabled nodes are required")
```
hardcoretime: Do not skip the test based on environment analysis. If this test case should be skipped, it must be done using an option in the run command.

Replace `Skip` with `Expect`.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+			}
+
+			masterNodes, err := listReadyNodes(ctx, f, map[string]string{kvmEnabledLabelKey: "true", masterLabelKey: "master"})
+			Expect(err).NotTo(HaveOccurred())
+			if len(masterNodes) == 0 {
+				Skip("at least one ready KVM-enabled master node is required")
```
hardcoretime: The same.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+			}
+		})
+
+		By("Creating vm-a", func() {
+			vmA = newPlacementVM("vm-a", f.Namespace().Name, nil)
+			err := f.CreateWithDeferredDeletion(ctx, newRootVDForPlacement(vmA.Name, f.Namespace().Name), vmA)
+			Expect(err).NotTo(HaveOccurred())
+
+			util.UntilObjectPhase(string(v1alpha2.MachineRunning), framework.LongTimeout, vmA)
+			util.UntilVMAgentReady(crclient.ObjectKeyFromObject(vmA), framework.LongTimeout)
```
hardcoretime: The guest agent status is not required in this test case.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+				newRootVDForPlacement(vmC.Name, f.Namespace().Name), vmC,
+				newRootVDForPlacement(vmD.Name, f.Namespace().Name), vmD,
+			}
+			err := f.CreateWithDeferredDeletion(ctx, objs...)
+			Expect(err).NotTo(HaveOccurred())
+
+			util.UntilObjectPhase(string(v1alpha2.MachineRunning), framework.LongTimeout, vmB, vmC, vmD)
+			util.UntilVMAgentReady(crclient.ObjectKeyFromObject(vmB), framework.LongTimeout)
+			util.UntilVMAgentReady(crclient.ObjectKeyFromObject(vmC), framework.LongTimeout)
+			util.UntilVMAgentReady(crclient.ObjectKeyFromObject(vmD), framework.LongTimeout)
```
hardcoretime: The same. This is not required and increases the test duration.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+			vmC = getVirtualMachine(ctx, f, vmC.Name)
+			vmD = getVirtualMachine(ctx, f, vmD.Name)
+
+			nodeA = vmA.Status.Node
+			Expect(nodeA).NotTo(BeEmpty())
+			Expect(vmB.Status.Node).NotTo(BeEmpty())
+			Expect(vmC.Status.Node).NotTo(BeEmpty())
+			Expect(vmD.Status.Node).NotTo(BeEmpty())
+
+			Expect(vmC.Status.Node).To(Equal(nodeA), "vm-a and vm-c should run on the same node")
```
hardcoretime: I think the current nodes of the virtual machines can be useful in the error message.

test/e2e/vm/affinity_toleration.go
```diff
@@ -0,0 +1,485 @@
...
+		vmbuilder.WithTolerations([]corev1.Toleration{{
+			Operator: corev1.TolerationOpExists,
+			Effect:   corev1.TaintEffectNoSchedule,
+		}}),
+	)
+	vm.Spec.Affinity = affinity
+	return vm
+}
+
+func newRootVDForPlacement(vmName, namespace string) *v1alpha2.VirtualDisk {
```
hardcoretime: Can `NewVDFromCVI` be used instead?

hardcoretime (changes requested):

hardcoretime (approved):

PR #2276, opened 2026-04-26
fix(module): extend virtualization RBACv1/RBACv2 coverage

fl64:
## Description

Extend virtualization RBAC coverage in both RBACv1 and RBACv2.                                               
                                                                                                                
   The change adds missing permissions for:                                                                     
   - `virtualmachinemacaddresses`                                                                               
   - `virtualmachinemacaddressleases`                                                                           
   - `virtualmachinesnapshotoperations`                                                                         
   - `nodeusbdevices`                                                                                           
                                                                                                                
   It also introduces a separate RBACv2 capability for snapshot operations:                                     
   - `d8:use:capability:virtualization:execute_virtualmachinesnapshot_operations`                               
                                                                                                                                 

Some virtualization API resources were missing from module RBAC templates, and RBACv1/RBACv2 coverage was inconsistent. 


<!---
  Describe steps to reproduce the expected result.
  What ACTION(s) to take to ensure the problem is gone.
-->


## Checklist
- [ ] e2e tests passed.


## Changelog entries
<!---
  /!\ See CONTRIBUTING.md for more details. /!\
  Examples:
  ```changes
  section: core
  type: feature
  summary: "Node restarts can be avoided by pinning a checksum to a node group in config values."
  ---
  section: core
  type: fix
  summary: "Nodes with outdated manifests are no longer provisioned on *InstanceClass update."
  impact_level: high
  impact: |
    Expect nodes of "Cloud" type to restart.
  ---
  impact_level: low
  ```
-->

```changes
   section: module                                                                                                             
   type: fix                                                                                                                   
   summary: "Added missing RBAC permissions for virtualization resources, including virtual machine MAC addresses, snapshot operations, and node USB devices."  
```


templates/user-authz-cluster-roles.yaml
```diff
@@ -19,8 +19,10 @@ rules:
   - virtualimages
   - virtualmachineblockdeviceattachments
   - virtualmachineipaddresses
+  - virtualmachinemacaddresses
   - virtualmachineclasses
   - virtualmachineoperations
+  - virtualmachinesnapshotoperations
   - usbdevices
```
Isteb4k, line 26: nodeusbdevices ?

> User: Has access to information about **all** objects

https://deckhouse.io/products/virtualization-platform/documentation/admin/platform-management/access-control/role-model.html#current-role-based-model

templates/user-authz-cluster-roles.yaml
```diff
@@ -47,6 +49,7 @@ rules:
   - virtualization.deckhouse.io
   resources:
   - virtualmachineoperations
+  - virtualmachinesnapshotoperations
```
Isteb4k: > PrivilegedUser: The same as User + can exec into containers, read secrets, and delete pods (and thus, restart them).

It looks like it should also be able to delete virtual machines.

templates/user-authz-cluster-roles.yaml
```diff
@@ -110,6 +115,8 @@ rules:
   - virtualization.deckhouse.io
   resources:
   - virtualmachineipaddressleases
+  - virtualmachinemacaddressleases
+  - nodeusbdevices
```
Isteb4k, line 104: May be move nodeusbdevices to regular User role

templates/user-authz-cluster-roles.yaml
Isteb4k: It doesn’t match the documentation at all. But I guess we can leave it like this for now

templates/user-authz-cluster-roles.yaml
```diff
@@ -147,16 +163,10 @@ rules:
   - virtualization.deckhouse.io
   resources:
   - virtualmachineipaddressleases
+  - virtualmachinemacaddressleases
```
Isteb4k, line 151: Just as we don't expect the user to create nodeusbdevices themselves and don't grant `create` permissions for that, we also don't expect the user to create virtualmachinemacaddressleases/virtualmachineipaddressleases. So we probably shouldn't grant `create` permissions for those either.

Or alternatively, grant `create` permissions for all three.

templates/user-authz-cluster-roles.yaml
Isteb4k: virtualmachinesnapshotoperations will allow the user to create virtual machines. But a PrivilegedUser shouldn’t be able to do that

templates/rbacv2/manage/permissions/manage_internals.yaml
```diff
@@ -42,6 +42,7 @@ rules:
   - virtualization.deckhouse.io
   resources:
   - virtualmachineipaddressleases
+  - virtualmachinemacaddressleases
```
Isteb4k, line 45: It seems like we shouldn't allow manual creation of virtualmachinemacaddressleases/virtualmachineipaddressleases - just like with nodeusbdevices.

Or alternatively, grant `create` permissions for all three.

Isteb4k (changes requested):

templates/rbacv2/manage/permissions/manage_internals.yaml
fl64: missed thing
  - create
removed

Isteb4k (approved):

templates/user-authz-cluster-roles.yaml
fl64: it doesn't fit into the logic of VM operation

fl64: yep
moved it to Editor and vmop too

PR #2278, opened 2026-04-27
fix(vmop): restore requested virtual disk size from snapshot metadata

danilrwx:
## Description
Restore the original requested VirtualDisk PVC size when restoring VM disks from snapshots.

The change stores `vd-original-size` in the VolumeSnapshot annotations during VirtualDiskSnapshot creation and uses this annotation when VMOP restores VirtualDisks from snapshot resources. If the annotation is absent, the restore flow keeps the previous behavior.

A restored VirtualDisk could be created without the original requested PVC size and rely on the snapshot restore defaults. This could make the restored disk size differ from the size requested by the source VirtualDisk.

Persisting and reading the original size from snapshot metadata makes VMOP restore recreate VirtualDisks with the same requested PVC size as the original disk.

1. Create a VirtualDisk with a requested PVC size.
2. Create a VirtualDiskSnapshot for this disk.
3. Restore the VM or disk via VMOP from the snapshot.
4. Verify that the restored VirtualDisk has `spec.persistentVolumeClaim.size` equal to the original requested size.
5. Verify that the corresponding VolumeSnapshot contains the `virtualization.deckhouse.io/vd-original-size` annotation.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: vd
type: fix
summary: Restore requested VirtualDisk size from snapshot metadata.
impact_level: low
```

### DIFF
--- a/images/virtualization-artifact/pkg/common/annotations/annotations.go
+++ b/images/virtualization-artifact/pkg/common/annotations/annotations.go
@@ -198,8 +198,8 @@ const (
 	AnnVirtualDiskOriginalAnnotations = AnnAPIGroupV + "/vd-original-annotations"
 	// AnnVirtualDiskOriginalLabels is the annotation for storing original VirtualDisk labels.
 	AnnVirtualDiskOriginalLabels = AnnAPIGroupV + "/vd-original-labels"
-	// AnnVirtualDiskRequestedSize is the annotation for storing original VirtualDisk requested PVC size.
-	AnnVirtualDiskRequestedSize = AnnAPIGroupV + "/vd-requested-size"
+	// AnnVirtualDiskOriginalSize is the annotation for storing original VirtualDisk requested PVC size.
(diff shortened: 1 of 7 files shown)


images/virtualization-artifact/pkg/common/annotations/annotations.go
```diff
@@ -198,6 +198,8 @@ const (
 	AnnVirtualDiskOriginalAnnotations = AnnAPIGroupV + "/vd-original-annotations"
 	// AnnVirtualDiskOriginalLabels is the annotation for storing original VirtualDisk labels.
 	AnnVirtualDiskOriginalLabels = AnnAPIGroupV + "/vd-original-labels"
+	// AnnVirtualDiskRequestedSize is the annotation for storing original VirtualDisk requested PVC size.
+	AnnVirtualDiskRequestedSize = AnnAPIGroupV + "/vd-requested-size"
```
diafour: May be `vd-original-size` as labels and annotations? Or `vd-original-requested-size`?

diafour (approved):

PR #2334, opened 2026-05-11
docs: add CRD examples

prismagod:
## Description
Added `x-examples` to CRD files.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: docs
type: docs
summary: "Added `x-examples` field to CRD files."
impact_level: low
```

### DIFF
--- a/crds/virtualmachineipaddresses.yaml
+++ b/crds/virtualmachineipaddresses.yaml
@@ -22,6 +22,14 @@ spec:
         openAPIV3Schema:
           description: |
             Defines the IP address for a virtual machine.
+          x-examples:
+            - apiVersion: virtualization.deckhouse.io/v1alpha2
+              kind: VirtualMachineIPAddress
+              metadata:
+                name: linux-vm-custom-ip
+              spec:
+                type: Static
+                staticIP: 10.66.20.77
           properties:
             apiVersion:
               type: string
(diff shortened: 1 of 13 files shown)


crds/virtualmachineipaddresses.yaml
```diff
@@ -22,6 +22,14 @@ spec:
         openAPIV3Schema:
           description: |
             Defines the IP address for a virtual machine.
+          x-examples:
+            - apiVersion: virtualization.deckhouse.io/v1alpha2
+              kind: VirtualMachineIPAddress
+              metadata:
+                name: linux-vm-custom-ip
+              spec:
```
Isteb4k, line 30: Let’s make it Auto instead of Static.

```suggestion
spec:
  type: Auto
```

crds/virtualmachinesnapshotoperations.yaml
```diff
@@ -44,6 +44,18 @@ spec:
...
+          x-examples:
+            - apiVersion: virtualization.deckhouse.io/v1alpha2
+              kind: VirtualMachineSnapshotOperation
+              metadata:
+                name: vmsop-name
+              spec:
+                type: CreateVirtualMachine
+                virtualMachineSnapshotName: vm-snapshot-clone
+                createVirtualMachine:
+                  mode: DryRun | Strict | BestEffort
```
Isteb4k: So should this example actually be a working example? The CRD already documents that mode has three options - DryRun | Strict | BestEffort. An example usually implies something runnable, doesn’t it? 

Empty nameReplacements and an empty customization also look a bit odd.

Isteb4k (changes requested):

Isteb4k (approved):

PR #2355, opened 2026-05-15
chore(ci): e2e report improve

universal-itengineer:
## Description

Improve the E2E messenger report generated by CI and reorganize the
report scripts so the same logic is easier to follow.

User-facing changes:

- Add a `Reason` column to the failed-tests thread table and rename
  `Test group` to `Tests`.
- Link cluster names in failed-tests thread replies to the matching
  workflow/job URL.
- Hide the `Errors` column in the main test-results table when there
  are no Ginkgo errors.
- Calculate `Success Rate` from executed specs only:
  `passed / (passed + failed + errors)`.
- Add table header emojis to the main report.
- Use the full Ginkgo `Failure.Message` text as the failed-test reason.
- Stop wrapping `*exec.ExitError` with `%w` in
  `test/e2e/internal/framework/ssh.go` so Ginkgo no longer dumps Go
  internals into the report when an SSH command fails.
- Capture Ginkgo stdout/stderr in the reusable pipeline and parse it as
  a fallback report source so suite-level failures such as
  `SynchronizedBeforeSuite` are still surfaced when the JSON report is
  missing or contains only setup failure data.

Internal refactors (no behavior change, covered by existing tests):

- Move the shared `withTempDir` / `createCore` jest helpers into
  `shared/test-utils.js` to remove duplication across three test files.
- Extract `escapeRegExp` in `shared/report-model.js`; consolidate
  `buildStatusMessage` into a status → template lookup table; remove a
  redundant ternary in `buildClusterStatus`.
- Unify the Ginkgo source lookup (`findGinkgoReport` /
  `findGinkgoOutput` → one `findGinkgoSource`) and call
  `parseGinkgoFile` once with the descriptor picked up front.
- Iterate over a list of required keys in
  `requireClusterReportConfig` instead of three identical `if` blocks.
- Pre-filter cluster reports once by storage key and merge
  `renderClusterFailuresSection` / `renderMissingReportsSection` into
  a single `renderBulletSection` helper. Output stays byte-identical;
  verified by existing `messenger-report` tests.
- Pass the `loop` credentials object as-is to `postToLoopApi` so the
  two call sites collapse to one-liners.

The current E2E notification thread only lists failed test groups and
does not show why a test failed. This forces engineers to open CI
artifacts and logs even for simple failures.

The main report also counted skipped specs in the `Success Rate`
denominator, which made Ginkgo runs look worse than the executed-spec
result. SSH command failures contained noisy Go internals in
`Failure.Message` because `*exec.ExitError` was wrapped with `%w`.

When E2E fails in `SynchronizedBeforeSuite`, Ginkgo may skip all specs
and the JSON report may be absent or contain no regular test failures.
Previously the messenger report could only show a generic storage
failure (for example `nfs` with `E2E TEST FAILED`), without the actual
setup failure reason. The pipeline now captures Ginkgo stdout into
`e2e_output_<storage>_<date>.log` and the report parses suite-level
failures out of it as a fallback.

The follow-up refactors remove duplicated helpers, collapse parallel
branches, and replace inline regex/`if`-chain logic with declarative
data, making the report scripts easier to read and extend without
changing the rendered output.

1. Run the E2E report generation with a Ginkgo JSON report containing
   failed specs.
2. Check the main messenger report:
   - the `Errors` column is hidden when all error counts are zero;
   - `Success Rate` is calculated without skipped specs;
   - table headers include emojis.
3. Check the failed-tests thread:
   - cluster names are links;
   - the table has `Tests` and `Reason` columns;
   - `Reason` contains the Ginkgo `Failure.Message` text.
4. SSH command failures are rendered as useful text instead of dumping
   `*exec.ExitError` internals.
5. If Ginkgo fails in `SynchronizedBeforeSuite` and no JSON report is
   created, the report is still generated from the captured Ginkgo
   output and the failed-tests thread shows the suite-level failure
   reason. Suite-level failures do not change spec metrics such as
   passed, failed, errors, or total.
6. The jest suite under `.github/scripts/js/e2e/report/` still passes
   (37 tests) and `tmp/test-ci/report/run.sh` produces the same
   markdown for the bundled fixtures.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: ci
type: fix
summary: Improve E2E messenger report failure reasons, suite setup failure reporting, and success-rate calculation.
impact_level: low
```

### DIFF
--- a/.github/scripts/js/e2e/report/messenger/markdown.js
+++ b/.github/scripts/js/e2e/report/messenger/markdown.js
@@ -217,21 +217,13 @@ function getFailedTestEntries(report) {
     return report.failedTestDetails.map((test) => ({
       name: test.name,
       reason: test.reason,
-      message: test.message,
     }));
   }
 
   return (report.failedTests || []).map((testName) => ({
     name: testName,
     reason: "",
-    message: "",
   }));
-}
-
-function sanitizeCodeBlock(value) {
-  return String(value || "")
-    .replace(/```/g, "`\u200b``")
-    .trim();
 }
 
 function summarizeFailedTestGroups(failedTestEntries) {
@@ -243,24 +235,16 @@ function summarizeFailedTestGroups(failedTestEntries) {
     if (!groups.has(groupName)) {
       groups.set(groupName, {
         reasons: new Set(),
-        details: [],
       });
     }
 
     const group = groups.get(groupName);
     group.reasons.add(reason);
-    if (test.message) {
-      group.details.push({
-        name: groupName,
-        message: test.message,
-      });
-    }
   }
 
   return Array.from(groups, ([name, group]) => ({
     name,
-    reason: Array.from(group.reasons).join("<br>"),
-    details: group.details,
+    reason: Array.from(group.reasons).join("; "),
   }));
 }
 
@@ -278,20 +262,6 @@ function renderFailedTestsThreadMessage(report) {
       lines.push(
         `| ${sanitizeCell(group.name)} | ${sanitizeCell(group.reason)} |`
       );
-    }
-
-    const details = failedGroups.flatMap((group) => group.details || []);
-    if (details.length > 0) {
-      lines.push("");
-      lines.push("**Details**");
-      for (const detail of details) {
-        lines.push("");
-        lines.push(`_${sanitizeListItem(detail.name)}_`);
-        lines.push("");
-        lines.push("```text");
-        lines.push(sanitizeCodeBlock(detail.message));
-        lines.push("```");
-      }
     }
   } else {
     lines.push(
--- a/.github/scripts/js/e2e/report/messenger/markdown.js
+++ b/.github/scripts/js/e2e/report/messenger/markdown.js
@@ -265,8 +265,7 @@ function summarizeFailedTestGroups(failedTestEntries) {
 }
 
 function renderFailedTestsThreadMessage(report) {
-  const clusterName = sanitizeListItem(report.cluster || report.storageType);
-  const lines = [`**${clusterName}**`];
+  const lines = [`**${formatClusterLink(report)}**`];
 
   if (Array.isArray(report.failedTests) && report.failedTests.length > 0) {
     const failedGroups = summarizeFailedTestGroups(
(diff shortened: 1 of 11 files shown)


.github/scripts/js/e2e/report/messenger/markdown.js
```diff
@@ -78,21 +78,37 @@ function renderTestResultsSection(testsReports) {
...
+          ? `| ${formatClusterLink(report)} | ${metrics.passed || 0} | ${
+              metrics.skipped || 0
+            } | ${metrics.failed || 0} | ${metrics.errors || 0} | ${
+              metrics.total || 0
+            } | ${formatRate(metrics.successRate)} |`
+          : `| ${formatClusterLink(report)} | ${metrics.passed || 0} | ${
+              metrics.skipped || 0
+            } | ${metrics.failed || 0} | ${metrics.total || 0} | ${formatRate(
+              metrics.successRate
+            )} |`
```
nevermarine: change this to if/else for clarity

.github/scripts/js/e2e/report/messenger/markdown.js
```diff
@@ -193,21 +209,59 @@ function getFailedTestGroupName(testName) {
...
 function renderFailedTestsThreadMessage(report) {
-  const clusterName = sanitizeListItem(report.cluster || report.storageType);
-  const lines = [`**${clusterName}**`];
+  const lines = [`**${formatClusterLink(report)}**`];
 
   if (Array.isArray(report.failedTests) && report.failedTests.length > 0) {
-    const failedGroups = summarizeFailedTestGroups(report.failedTests);
+    const failedGroups = summarizeFailedTestGroups(
+      getFailedTestEntries(report)
+    );
```
nevermarine: maybe call getFailedTestEntries from inside summarizeFailedTestGroups?

nevermarine (changes requested):

.github/scripts/js/e2e/report/messenger/markdown.js
universal-itengineer: fixed

.github/scripts/js/e2e/report/messenger/markdown.js
universal-itengineer: fixed

nevermarine (approved):

PR #2406, opened 2026-05-26
chore(vm): more messages about cpu and memory hotplug fails

diafour:
## Description

- Do not start live migration to apply cpu or memory changes with hotplug if VM is NonMigratable.
- Fix sync_kvvm_test: fill PrintableStatus for kvvm as it is required to detect if disruptive changes may be applied.
- Add test for Upgrade*ChangesToRestart mutators.

No need to run live migration if VM is NonMigratable, backoff to RequireRestart.

CPU and memory changes are always require restart if VM in NonMigratable.

## Checklist
- [ ] e2e tests passed.


LopatinDmitr (commented):
Please add tests for negative cases

LopatinDmitr (dismissed):

LopatinDmitr (approved):

PR #2407, opened 2026-05-26
fix(vm): allow disk updates when VM is stopped

hardcoretime:
## Description

Fixed the issue where disk changes on a stopped VirtualMachine with WaitForFirstConsumer (wffc) volume binding mode were not being applied to the KVVM (intvirtvm) resource.

Changes made to `pkg/controller/kvbuilder/kvvm_utils.go`:
1. Added `isVmRunning` check to the paravirtualization condition in `applyBlockDeviceRefs` to allow new disks when VM is stopped
2. Updated `cleanupRemovedStaticDisks` function to accept `isVmRunning` parameter and remove all disks unconditionally when VM is stopped (not just non-hotpluggable ones)
3. Added VMBDA filtering to prevent removing disks that are attached via VirtualMachineBlockDeviceAttachment - these disks should remain in KVVM spec to be visible in stopped VM status

Added unit tests in `pkg/controller/kvbuilder/kvvm_utils_test.go` to verify:
- When VM is stopped: all disks not in VM spec are removed (regardless of hotpluggable flag)
- When VM is running: only non-hotpluggable disks not in VM spec are removed
- VMBDA-attached disks are preserved even when VM is stopped

The DataExport e2e test was failing when using storage class with WaitForFirstConsumer (wffc) volume binding mode. The scenario:
1. Create a VM with one disk (vd-data)
2. Stop the VM
3. Run DataExport which:
   - Removes the original disk (vd-data)
   - Creates new disks from snapshot/source (vd-restored-from-disk, vd-restored-from-snapshot)
4. Start the VM

The VM got stuck in starting phase because:
- Old disk (vd-data) was NOT removed from KVVM (because it was marked as hotpluggable)
- New restored disks were NOT added to KVVM (because paravirtualization was enabled and VM had existing volumes)

This happened because the code was designed for running VMs with hotplug handler, not for stopped VMs where changes should be applied directly.

When a VM is stopped and disks are changed in the VM spec (disks added/removed), the KVVM resource should be updated accordingly. After the fix:
- Old disks not in spec are removed from KVVM (when VM is stopped)
- New disks in spec are added to KVVM (when VM is stopped)
- For running VMs, the behavior remains unchanged (hotplug handler manages disk changes)
- Disks attached via VMBDA are preserved in KVVM spec to remain visible in stopped VM status

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: "Fixed disk update for stopped VMs with WaitForFirstConsumer storage class. Previously, when a VM was stopped and disks were changed, the KVVM was not updated, causing the VM to get stuck in starting phase."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
+++ b/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
@@ -97,6 +97,7 @@ func ApplyVirtualMachineSpec(
 	vdByName map[string]*v1alpha2.VirtualDisk,
 	viByName map[string]*v1alpha2.VirtualImage,
 	cviByName map[string]*v1alpha2.ClusterVirtualImage,
+	vmbdaByBlockDeviceRef map[v1alpha2.VMBDAObjectRef][]*v1alpha2.VirtualMachineBlockDeviceAttachment,
 	class *v1alpha2.VirtualMachineClass,
 	ipAddress string,
 	networkSpec network.InterfaceSpecList,
@@ -130,7 +131,7 @@ func ApplyVirtualMachineSpec(
 		return err
 	}
 
-	if err := applyBlockDeviceRefs(kvvm, vm, isVmRunning, vdByName, viByName, cviByName); err != nil {
+	if err := applyBlockDeviceRefs(kvvm, vm, isVmRunning, vdByName, viByName, cviByName, vmbdaByBlockDeviceRef); err != nil {
 		return err
 	}
 
@@ -166,6 +167,7 @@ func applyBlockDeviceRefs(
 	vdByName map[string]*v1alpha2.VirtualDisk,
 	viByName map[string]*v1alpha2.VirtualImage,
 	cviByName map[string]*v1alpha2.ClusterVirtualImage,
+	vmbdaByBlockDeviceRef map[v1alpha2.VMBDAObjectRef][]*v1alpha2.VirtualMachineBlockDeviceAttachment,
 ) error {
 	hasExplicitBootOrder := false
 	for _, bd := range vm.Spec.BlockDeviceRefs {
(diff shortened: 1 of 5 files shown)


deckhouse-BOaTswain:
Workflow has started.
  Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/26497982167)

  The target step completed with status: **failure**.

deckhouse-BOaTswain:
Workflow has started.
  Follow the progress here: [Workflow Run](https://github.com/deckhouse/virtualization/actions/runs/26536501205)

  The target step completed with status: **failure**.

images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils_test.go
```diff
@@ -0,0 +1,143 @@
...
+				newDisk1Name: {},
+			}
+			hotpluggableVolumes := map[string]struct{}{}
+
+			// Simulate disk attached via VMBDA
+			vmbdaDiskNames := map[string]struct{}{
+				oldDisk1Name: {},
+			}
+
+			cleanupRemovedStaticDisks(kvvm, specDiskNames, hotpluggableVolumes, vmbdaDiskNames, false)
```
loktev-d: describe says isVmRunning=true but it's false here

images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
```diff
@@ -185,13 +187,22 @@ func applyBlockDeviceRefs(
 			hotpluggableVolumes[v.Name] = struct{}{}
 		}
 	}
+	vmbdaDiskNames := make(map[string]struct{}, len(vmbdaByBlockDeviceRef))
+	for ref := range vmbdaByBlockDeviceRef {
```
loktev-d, line 191: probably should also check vmbda status here, adding only Attached

hardcoretime: When the virtual machine is stopped, its VMBDA is in the Pending phase. In this case, the controller will remove the block device from the KVVM specification.

loktev-d (approved):

PR #2414, opened 2026-05-28
refactor(ci): workflows nested cluster

universal-itengineer:
## Description
Refactor the nested-cluster E2E CI workflows and supporting automation.

This PR replaces large inline workflow shell blocks with versioned scripts under `.github/scripts/bash/e2e`, introduces reusable composite actions for common CI steps, and splits the previous nested-cluster workflow layout into dedicated nightly and release-oriented pipelines. The reusable pieces now cover run ID generation, toolchain setup, `d8` installation with retry/backoff, nested kubeconfig setup, private registry login, encrypted artifact upload/help text, storage configuration, virtualization configuration, wait helpers, VM power-off, image digest verification, and release test execution.

It also adds direct lint coverage for GitHub Actions and CI shell scripts, moves the GitHub Actions e2e runner into `test/e2e/scripts/e2e-ci.sh`, and exposes it as `task e2e:run:ci` from the repository root. The static nested-cluster values rendering is unified through `test/dvp-static-cluster/values.yaml.tmpl` and `.github/scripts/bash/e2e/render-dvp-static-values.sh`; runtime differences such as enabled modules, Kubernetes version, worker memory, and additional disk size are passed as workflow inputs or environment variables instead of being duplicated in workflow shell blocks.

The old nested-cluster E2E workflows mixed orchestration with long inline shell logic. That made changes difficult to review, duplicated nightly and release behavior, and left important CI paths without focused linting or reusable entrypoints.

Moving operational logic into scripts and composite actions makes the pipelines easier to review, lint, test locally, and reuse between nightly and release scenarios. The shared values template removes drift between cluster configurations, while the smaller workflow files make release-specific behavior explicit through inputs rather than copied shell fragments.

Nested-cluster nightly and release E2E jobs continue to provision prerequisites, render static cluster values, configure storage and virtualization, run the required Ginkgo suites, collect encrypted artifacts, verify deployed image digests where needed, and clean up or preserve resources according to the workflow phase.

Validation for these changes is available with:

- `task lint:shellcheck`
- `task lint:actionlint`
- `task e2e:run:ci` from the repository root
- `task run:ci` inside `test/e2e`

## Checklist
- [ ] The code is covered by unit tests. Not applicable: this is CI workflow and scripting refactoring.
- [ ] e2e tests passed.
- [x] Documentation updated according to the changes. Composite action README files were added.

## Changelog entries
```changes
section: ci
type: chore
summary: Refactor nested-cluster E2E workflows into reusable pipelines, scripts, and composite actions.
impact_level: low
```


.github/actions/registry-login/action.yml
```diff
@@ -0,0 +1,28 @@
...
+      env:
+        DOCKER_CFG: ${{ inputs.docker_cfg }}
+      run: |
+        REGISTRY=$(base64 -d <<< "$DOCKER_CFG" | jq '.auths | to_entries | .[] | .key' -r)
+        USERNAME=$(base64 -d <<< "$DOCKER_CFG" | jq -r '.auths | to_entries | .[] | .value.auth' | base64 -d | cut -d ':' -f1)
+        PASSWORD=$(base64 -d <<< "$DOCKER_CFG" | jq -r '.auths | to_entries | .[] | .value.auth' | base64 -d | cut -d ':' -f2)
+
+        echo "::add-mask::$USERNAME"
+        echo "::add-mask::$PASSWORD"
+        echo "$PASSWORD" | docker login "$REGISTRY" --username "$USERNAME" --password-stdin
```
nevermarine, line 26: need to test this

.github/scripts/bash/e2e/configure-virtualization.sh
```diff
@@ -8,9 +8,26 @@ source "${SCRIPT_DIR}/common.sh"
...
+required_env_value() {
+  local name="$1"
+
+  require_env "$name"
+  printf '%s' "${!name}"
+}
+
+dev_registry_docker_cfg="$(required_env_value DEV_REGISTRY_DOCKER_CFG)"
+nested_storage_class_name="$(required_env_value NESTED_STORAGE_CLASS_NAME)"
+virtualization_tag="$(required_env_value VIRTUALIZATION_TAG)"
```
nevermarine: maybe rewrite this so there would be only one function `required_env`?

.github/workflows/e2e-test-releases-reusable-pipeline.yml
nevermarine: can't we use run-release-e2e.sh here as well?

test/e2e/Taskfile.yaml
```diff
@@ -44,6 +44,15 @@ tasks:
...
 
+  e2e:ci:
+    desc: "Run e2e tests in GitHub Actions"
+    deps:
+      - copy
+      - kubectl
+      - d8
+    cmds:
+      - bash ./scripts/e2e-ci.sh
+
```
nevermarine: wouldn't this result in duplication of e2e when running from the root of the repo? i.e., `task e2e:e2e:ci`?

nevermarine (changes requested):

.github/scripts/bash/e2e/configure-virtualization.sh
universal-itengineer: fixed

test/e2e/Taskfile.yaml
universal-itengineer: fixed

.github/workflows/e2e-test-releases-reusable-pipeline.yml
universal-itengineer: fixed

.github/actions/registry-login/action.yml
universal-itengineer: The masking + login is correct by GitHub's docs:

- The decoded username/password are freshly derived strings, so they aren't auto-masked — the explicit ::add-mask:: (lines 24-25) is required and is placed before the only command that uses them (docker login).
- --password-stdin keeps the password out of the process args.
- Steps run as bash --noprofile --norc -eo pipefail, so a broken base64 / missing auth will fail the step.


Masking with ::add-mask:: (manually masking values derived at runtime, like the decoded username/password):
    - https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#masking-a-value-in-a-log
Note that auto-masking applies only to registered secrets: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#about-secrets

Default bash shell flags -eo pipefail (and --noprofile --norc):
       - https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell (see the "bash" row in the default-shell table / exit-code handling)

nevermarine (approved):

PR #2417, opened 2026-05-29
test: add postcleanup mode

yaroslavborbat:
## Description
Replace boolean `IsCleanupNeeded` flag with `PostCleanupMode` enum in e2e test config.

New values for `POST_CLEANUP` env:
- `always` (default) — always clean up resources after tests
- `never` — never clean up (useful for debugging)
- `no-on-failure` — skip cleanup only when the test suite fails, preserving resources on failure for investigation (cleanup runs on success)

The old boolean flag (`yes`/`no`) didn't allow keeping resources when tests fail while still cleaning up on success. The new `no-on-failure` mode fills this gap, making it easier to debug failing e2e tests without manually toggling cleanup.

1. Set `POST_CLEANUP=no-failure` before running e2e tests
2. If tests pass — resources are cleaned up as usual
3. If tests fail — resources remain in the cluster for investigation

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: test
type: chore
summary: Replace boolean POST_CLEANUP flag with PostCleanupMode enum (always/never/no-on-failure) in e2e config.
impact_level: low
```

### DIFF
--- a/test/e2e/internal/config/cleanup.go
+++ b/test/e2e/internal/config/cleanup.go
@@ -16,6 +16,11 @@ limitations under the License.
 
 package config
 
+import (
+	"fmt"
+	"os"
+)
+
 // PrecreatedCVICleanupEnv defines an environment variable to explicitly enable deletion of precreated CVIs after the suite.
 //
 // By default, precreated CVIs are not deleted: they are shared across runs and may be reused.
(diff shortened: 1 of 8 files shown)


test/e2e/internal/config/cleanup.go
```diff
@@ -16,12 +16,23 @@ limitations under the License.
...
+
 // PrecreatedCVICleanupEnv defines an environment variable to explicitly enable deletion of precreated CVIs after the suite.
 //
 // By default, precreated CVIs are not deleted: they are shared across runs and may be reused.
 // Set PRECREATED_CVI_CLEANUP=yes to delete them after the suite; unset, empty, or "no" means no deletion.
 const PrecreatedCVICleanupEnv = "PRECREATED_CVI_CLEANUP"
 
 // PostCleanupEnv defines an environment variable used to explicitly request the deletion of created/used resources.
-// Valid values: "yes", "no", or "" (default = yes).
+// Valid values: "always", "never", "no-failure", or "" (default = always).
```
hardcoretime:
```suggestion
// Valid values: "always", "never", "no-on-failure", or "" (default = always).
```

hardcoretime (changes requested):

hardcoretime (approved):

PR #2419, opened 2026-06-01
feat(vm): fallback hotplug resource updates to restart on quota limits

eofff:
## Description

Added a runtime check for VM CPU/memory hotplug changes that require live migration. If the project `ResourceQuota` cannot fit the temporary target pod allocation needed during hotplug migration, the VM controller now keeps the VM update accepted but marks the change as requiring restart instead of trying to apply it live.

The workload updater also skips creating `hotplug-resources-*` migration operations when the VM is already waiting for restart to apply configuration changes.

Hotplug CPU/memory updates can require a live migration, which temporarily needs additional CPU/memory quota for the migration target pod. When that extra quota is unavailable, starting migration is pointless: it can fail or migrate without applying the requested resource change clearly.

Admission rejection is not the right behavior here because the user should still be able to update the VM spec. Instead, the controller reports that a restart is required and includes a quota-specific reason in `AwaitingRestartToApplyConfiguration`.

When hotplug CPU/memory changes are requested:
- If migration-time quota is sufficient, hotplug migration proceeds as before.
- If migration-time quota is insufficient, the VM enters `AwaitingRestartToApplyConfiguration=True` and the condition message explains which quota/resource is insufficient.
- The workload updater does not create a hotplug migration VMOP while the VM is waiting for restart.

Example condition message:

```text
Hotplug migration cannot start because project quota "all-pods" has insufficient requests.cpu: required additional 10, available 3. Restart the virtual machine to apply the changes without live migration.
```

## Checklist
- [x] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: "Fallback CPU/memory hotplug updates to restart when project quota cannot fit migration-time resources."
```


images/virtualization-artifact/pkg/controller/vm/internal/validators/hotplug_resources_validator.go
```diff
@@ -0,0 +1,157 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/deckhouse/virtualization-controller/pkg/common"
+	"github.com/deckhouse/virtualization-controller/pkg/controller/kvbuilder"
+	"github.com/deckhouse/virtualization/api/core/v1alpha2"
+)
+
+const maxHotplugCores = 128
+
+var maxHotplugMemory = resource.MustParse("256Gi")
```
diafour: These constants are already exist in kv_builder package.

images/virtualization-artifact/pkg/controller/vm/internal/validators/hotplug_resources_validator.go
```diff
@@ -0,0 +1,157 @@
...
+	"k8s.io/apimachinery/pkg/api/resource"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/deckhouse/virtualization-controller/pkg/common"
+	"github.com/deckhouse/virtualization-controller/pkg/controller/kvbuilder"
+	"github.com/deckhouse/virtualization/api/core/v1alpha2"
+)
+
+const maxHotplugCores = 128
```
diafour: CPUCountValidator already has "max cores" validation.

images/virtualization-artifact/pkg/controller/vm/internal/validators/hotplug_resources_validator.go
```diff
@@ -0,0 +1,157 @@
...
+	if oldVM.Spec.CPU.CoreFraction != newVM.Spec.CPU.CoreFraction {
+		return true
+	}
+	return oldVM.Spec.Memory.Size.Cmp(newVM.Spec.Memory.Size) != common.CmpEqual
+}
+
+func validateHotplugRanges(vm *v1alpha2.VirtualMachine) error {
+	if vm.Spec.CPU.Cores > maxHotplugCores {
+		return fmt.Errorf("hotplug CPU cores should not exceed %d", maxHotplugCores)
+	}
```
diafour: CPUCountValidator already has "max cores" validation.

images/virtualization-artifact/pkg/controller/vm/internal/validators/hotplug_resources_validator.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if vm.Spec.Memory.Size.Cmp(maxHotplugMemory) == common.CmpGreater {
+		return fmt.Errorf("hotplug memory should not exceed %s", maxHotplugMemory.String())
```
diafour: Changing memory size is not always "hotplug", e.g. changing from 256Mi to 256Gi requires VM reboot without live migration.
Make message more neutral:

```suggestion
		return fmt.Errorf("memory size should not exceed %s", maxHotplugMemory.String())
```

images/virtualization-artifact/pkg/controller/vm/internal/validators/hotplug_resources_validator.go
```diff
@@ -0,0 +1,157 @@
...
+
+	used := quota.Status.Used[resourceName]
+	free := hard.DeepCopy()
+	free.Sub(used)
+	if free.Sign() < 0 {
+		free.Set(0)
+	}
+
+	if newReq.Cmp(hard) == common.CmpGreater {
+		return fmt.Errorf("hotplug %s request %s exceeds project quota %q hard limit %s", resourceName, newReq.String(), quota.GetName(), hard.String())
```
diafour: Same for validating memory size. Make message more neutral:


```suggestion
		return fmt.Errorf("%s request %s exceeds project quota %q hard limit %s", resourceName, newReq.String(), quota.GetName(), hard.String())
```

diafour (changes requested):
Also:

- changing memory.size and cpu.cores/cpu.coreFraction is not always a "hotplug", it may require reboot. Do we need this validation?
- It seems these validations are needed to prevent migration fails, so it only applicable for specific phases, e.g. no need to validate "Stopped" VMs.

images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
```diff
@@ -835,6 +850,102 @@ func (h *SyncKvvmHandler) isVMNonMigratable(
...
+
+	for i := range quotaList.Items {
+		quota := &quotaList.Items[i]
+		if message, allowed := quotaAllowsHotplugMigration(quota, corev1.ResourceRequestsCPU, newCPU); !allowed {
+			return message, true, nil
+		}
+		if message, allowed := quotaAllowsHotplugMigration(quota, corev1.ResourceRequestsMemory, newMemory); !allowed {
+			return message, true, nil
+		}
+	}
```
diafour, line 884: Let's check both quaots and return joined message if both cpu and memory are insufficient.

diafour (dismissed):

diafour (approved):

PR #2432, opened 2026-06-01
fix(core): add libvirt patch for USB startupPolicy in containers

yaroslavborbat:
## Description

Add a libvirt patch (`005-usb-startup-policy-in-containers.patch`) that fixes `startupPolicy='optional'` for USB hostdev in containerized environments.

The patch adds a `virFileExists` check on the device node path in `virUSBDeviceSearch` after sysfs discovery. If the `/dev/bus/usb/` node is missing, the device is skipped so that callers see an empty result and handle the absence gracefully.

See also:
- https://github.com/deckhouse/3p-kubevirt/pull/124
- https://github.com/containerd/containerd/blob/c8802b6e7b81b21ffc134a9c1028f472f87dc262/pkg/oci/mounts.go#L58

In containers, sysfs (`/sys/bus/usb/devices/`) is mounted from the host kernel and exposes all USB devices regardless of mount namespace isolation. This causes libvirt's `virUSBDeviceSearch` to report a device as available even when the corresponding `/dev/bus/usb/` node is not present in the container's mount namespace.

As a result, `startupPolicy='optional'` does not remove the missing USB hostdev from the domain XML during incoming migration, and QEMU fails to start with:

```
Could not open '/dev/bus/usb/005/002': No such file or directory
```

This breaks VM live migration when USB devices are attached via hotplug and the target node does not yet have the device node mounted into the virt-launcher pod.

1. Start a VM with a hotplugged USB device on node A.
2. Migrate the VM to node B where the USB device node is not yet in the pod's mount namespace.
3. The migration succeeds — libvirt skips the missing USB hostdev due to `startupPolicy='optional'`.
4. After migration, the USB device is hotplugged into the VM on the target node.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: core
type: fix
summary: Fix USB startupPolicy='optional' in containers where sysfs exposes host devices but device nodes are absent from the mount namespace.
impact_level: low
```

### DIFF
--- a/images/packages/libvirt/patches/005-usb-startup-policy-in-containers.patch
+++ b/images/packages/libvirt/patches/005-usb-startup-policy-in-containers.patch
@@ -1,0 +1,29 @@
+diff --git a/src/util/virusb.c b/src/util/virusb.c
+index cf3461f80a..5a92098452 100644
+--- a/src/util/virusb.c
++++ b/src/util/virusb.c
+@@ -169,6 +169,24 @@ virUSBDeviceSearch(unsigned int vendor,
+         if (!usb)
+             goto cleanup;
+ 
++        /* In containerized environments sysfs may expose USB devices
++         * from the host kernel while the corresponding device node
++         * under /dev/bus/usb/ is not present in the container's mount
++         * namespace.  Skip such devices so that callers relying on
++         * startupPolicy='optional' see an empty result and can handle
++         * the missing device gracefully. */
++        if (virUSBDeviceGetPath(usb) &&
++            !virFileExists(virUSBDeviceGetPath(usb))) {
++            VIR_DEBUG("USB device %03u:%03u found in sysfs but "
++                      "device node '%s' is not accessible, skipping",
++                      found_bus, found_devno,
++                      virUSBDeviceGetPath(usb));
++            g_clear_pointer(&usb, virUSBDeviceFree);
++            if (found)
++                break;
++            continue;
++        }
++
+         if (virUSBDeviceListAdd(list, &usb) < 0)
+             goto cleanup;
+ 
--- a/build/components/versions.yml
+++ b/build/components/versions.yml
@@ -3,7 +3,7 @@ firmware:
   libvirt: v10.9.0
   edk2: stable202411
 core:
-  3p-kubevirt: v1.6.2-v12n.41
+  3p-kubevirt: v1.6.2-v12n.42
   3p-containerized-data-importer: v1.60.3-v12n.19
   distribution: 2.8.3
 package:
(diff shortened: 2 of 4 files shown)


images/packages/libvirt/patches/005-usb-startup-policy-in-containers.patch
```diff
@@ -0,0 +1,29 @@
...
++         * namespace.  Skip such devices so that callers relying on
++         * startupPolicy='optional' see an empty result and can handle
++         * the missing device gracefully. */
++        if (virUSBDeviceGetPath(usb) &&
++            !virFileExists(virUSBDeviceGetPath(usb))) {
++            VIR_DEBUG("USB device %03u:%03u found in sysfs but "
++                      "device node '%s' is not accessible, skipping",
++                      found_bus, found_devno,
++                      virUSBDeviceGetPath(usb));
++            g_clear_pointer(&usb, virUSBDeviceFree);
```
diafour, line 21: `usb` is wrapped with `g_autoptr`, no need to cleanup explicitly. It will be cleaned automatically when `usb` goes out of scope, e.g. on return.

images/packages/libvirt/patches/005-usb-startup-policy-in-containers.patch
```diff
@@ -0,0 +1,29 @@
...
++            if (found)
++                break;
++            continue;
```
diafour, line 24: Is it possible to continue? We found device in sysfs, but device is not usable. It seems an unrecoverable error, so cleanup is needed:


```suggestion
+            goto cleanup;
```

yaroslavborbat: `goto cleanup` is not appropriate here because a missing device node in `/dev/bus/usb` is not treated as a fatal search error in this code path. We did find a matching entry in `sysfs`, but this extra check verifies that the device is actually usable from the current mount namespace. If the node is missing, that means the device must be ignored, not that the search itself failed.

  `virUSBDeviceSearch()` is used in two modes:
  - by `vendor/product`, where it returns a list and there may be multiple matches;
  - by `bus/devno`, where at most one match is expected.

  That is why the control flow differs:
  - `continue` is needed for `vendor/product` search, so we skip the inaccessible device and keep looking for other usable matches;
  - `if (found) break;` is needed for `bus/devno` search: once the matching `sysfs` entry was found but proved unusable, there is nothing else to search for, so we stop and return an empty result.

  Using `goto cleanup` here would turn this case into an error return instead of an empty result. That would break the `mandatory = false` semantics in the callers: they should observe this case as "device not found/usable", not as an internal error.

images/packages/libvirt/patches/005-usb-startup-policy-in-containers.patch
yaroslavborbat: `g_autoptr` only frees `usb` when it goes out of scope, i.e. at function exit. Here we need an explicit cleanup because this branch uses `continue`.

  Without `g_clear_pointer(&usb, virUSBDeviceFree)`, the loop would continue and the next assignment to `usb` would overwrite the current object, leaking it.

diafour (dismissed):

diafour (approved):

PR #2433, opened 2026-06-02
fix(vm): handle FailedMapVolume in pod volume error detection

eofff:
## Description
Added handling for one more pod volume warning reason (`FailedMapVolume`) and unified volume-error reason checks in the VM controller.

Changes:
- added `ReasonFailedMapVolume` constant in `volumeevent_watcher.go`;
- introduced `IsVolumeErrorReason(reason string)` helper to keep reason filtering in one place;
- replaced duplicated checks in watcher predicates/handlers and in VM lifecycle error extraction with the new helper.

Previously, VM pod volume errors were recognized only for `FailedAttachVolume` and `FailedMount`.
In some failure scenarios Kubernetes emits `FailedMapVolume`, so the controller could miss relevant warning events and return less informative VM error status.
After this change, all key volume-related warning reasons are processed consistently, so users get clearer and more complete diagnostics when VM startup fails due to volume issues.

1. Create/reproduce a VM startup issue where pod events include `FailedMapVolume` (or existing `FailedAttachVolume` / `FailedMount`).
2. Verify that the event is picked up by volume event watcher.
3. Verify that VM status/lifecycle returns a `VMPodVolumeError` with the correct `Reason` and `Message`.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: vm
type: fix
summary: "VM pod volume error handling now includes FailedMapVolume and surfaces more complete pod volume diagnostics."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/lifecycle_hotplug_pod_errors_test.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/lifecycle_hotplug_pod_errors_test.go
@@ -73,6 +73,25 @@ var _ = Describe("LifeCycleHandler hotplug pod errors", func() {
 			LastTimestamp: metav1.NewTime(time.Now()),
 		}
 	}
+	newKVVMIWithHotplugPod := func(vm *v1alpha2.VirtualMachine, pod *corev1.Pod, uid types.UID) *virtv1.VirtualMachineInstance {
+		return &virtv1.VirtualMachineInstance{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      vm.Name,
+				Namespace: vm.Namespace,
+			},
+			Status: virtv1.VirtualMachineInstanceStatus{
+				VolumeStatus: []virtv1.VolumeStatus{
+					{
+						Name: "vd-hotplug",
+						HotplugVolume: &virtv1.HotplugVolumeStatus{
+							AttachPodName: pod.Name,
+							AttachPodUID:  uid,
+						},
+					},
+				},
+			},
+		}
+	}
 
 	It("should return volume errors for launcher pod label", func() {
 		vm := &v1alpha2.VirtualMachine{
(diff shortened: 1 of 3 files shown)


images/virtualization-artifact/pkg/controller/vm/internal/watcher/volumeevent_watcher.go
```diff
@@ -48,6 +58,46 @@ type VolumeEventWatcher struct {
...
+	}
+
+	if vmName, hasLabel := pod.GetLabels()[virtv1.VirtualMachineNameLabel]; hasLabel {
+		return vmName, true
+	}
+
+	return "", false
+}
+
+func (w *VolumeEventWatcher) resolveVMNameByHotplugStatus(ctx context.Context, pod *corev1.Pod) (string, bool) {
```
LopatinDmitr: May be create one function?

images/virtualization-artifact/pkg/controller/vm/internal/watcher/volumeevent_watcher.go
```diff
@@ -48,6 +58,46 @@ type VolumeEventWatcher struct {
 	client client.Client
 }
 
+func getVirtualMachineNameFromPodLabels(pod *corev1.Pod) (string, bool) {
```
LopatinDmitr: Why return bool?

images/virtualization-artifact/pkg/controller/vm/internal/lifecycle_hotplug_pod_errors_test.go
```diff
@@ -0,0 +1,142 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	virtv1 "kubevirt.io/api/core/v1"
+
+	"github.com/deckhouse/virtualization-controller/pkg/common/testutil"
+	"github.com/deckhouse/virtualization-controller/pkg/controller/vm/internal/watcher"
+	"github.com/deckhouse/virtualization/api/core/v1alpha2"
+)
+
+var _ = Describe("LifeCycleHandler hotplug pod errors", func() {
```
LopatinDmitr, line 37: What would I add minimally?                                                                                                                                   
                                                                                                                                                               
 If you don't inflate the tests, I would add 3 cases.:                                                                                                                
                                                                                                                                                               
 ### 1. UID mismatch                                                                                                                                           
                                                                                                                                                               
 ```text                                                                                                                                                       
   should ignore hotplug pod when UID does not match KVVMI status                                                                                              
 ```                                                                                                                                                           
                                                                                                                                                               
 ### 2. Not ContainerCreating                                                                                                                                  
                                                                                                                                                               
 ```text                                                                                                                                                       
   should ignore volume event when pod is not ContainerCreating                                                                                                
 ```                                                                                                                                                           
                                                                                                                                                               
 ### 3. Non-volume event                                                                                                                                       
                                                                                                                                                               
 ```text                                                                                                                                                       
   should ignore non-volume warning events                                                                                                                     
 ```                                                                                                                                                           
                                                                                                                                                               
 These three tests greatly increase confidence.

LopatinDmitr (approved):

PR #2442, opened 2026-06-03
fix(cli): use local ssh client for vm access

LopatinDmitr:
## Description

Drop the embedded Go SSH/SCP implementation and run the local `ssh`/`scp`
binaries directly through a `ProxyCommand` (the `d8` port-forward subcommand).
Provide a uniform way to pass extra options to the local client and to run
a command on the VM.

Highlights of the user-facing surface:

- Single `--ssh-args` flag for the local `ssh`/`scp` client. Can be repeated
  or given as a single value with space-separated options:
  ```bash
  d8 ssh user@myvm --ssh-args='-o StrictHostKeyChecking=no -o LogLevel=ERROR'
  d8 ssh user@myvm --ssh-args='-o X' --ssh-args='-o Y'
  ```
- `--` separator for the command to execute on the VM (mirrors plain `ssh`):
  ```bash
  d8 ssh user@myvm -- 'ls -la /'
  d8 ssh user@myvm -- bash -lc 'echo $HOSTNAME'
  ```
- `-c`/`--command` remains supported as an equivalent to `--`.
- `--known-hosts` is mapped to OpenSSH `UserKnownHostsFile`; `--identity-file`
  to `-i`.
- `--local-ssh` and `--local-ssh-opts` are kept as deprecated compatibility
  flags (`--local-ssh-opts` is the old name for `--ssh-args`).
- `templates.MinimumArgs` was added so commands can consume everything
  after `--`.

The previous `virtctl`-style command shipped its own native Go SSH/SCP
client, including terminal handling, `knownhosts` parsing and `stdio`
tunnel plumbing. Maintaining that implementation:

- duplicated what OpenSSH already does well (proxy commands, keep-alives,
  host key verification, compression, control master, etc.);
- made it impossible to use features the local client already supports
  (`-L`, `-R`, `-D`, jump hosts, custom config in `~/.ssh/config`,
  `ProxyJump`, etc.) without growing the wrapper with more flags;
- meant every fix for an edge case (e.g. non-ASCII terminals on Windows)
  had to be ported to the embedded code.

Switching to the local `ssh`/`scp` client makes the CLI delegate to the
tooling users (and CI) already have, with a much smaller surface in Go
code. The new `--ssh-args` design replaces the previous `--ssh-opts` /
positional-tail approach, which was both verbose (each `-o` had to be
wrapped in its own flag) and inconsistent with the standard `ssh` CLI.

1. `d8 ssh user@myvm` opens an interactive SSH session through
   `d8 v port-forward --stdio=true` (ProxyCommand).
2. `d8 ssh user@myvm -- 'cmd'` runs `cmd` on the VM and exits with its
   status code; `d8 ssh user@myvm -c 'cmd'` does the same.
3. `d8 ssh user@myvm --ssh-args='-o X -o Y'` passes `-o X -o Y` to the
   local `ssh` (split into separate argv entries by whitespace).
4. `d8 scp local user@myvm:remote` copies a file via local `scp`; `scp`
   flags can be supplied via the same `--ssh-args` flag.
5. The deprecated `--local-ssh-opts` flag still works and emits a
   warning pointing to `--ssh-args`.

Manual smoke test (13 cases covering `--ssh-args` single/repeat/mixed
forms, `--` vs `-c`, deprecation warning, scp up- and download) was run
against `vm-alpine` in the `test` namespace.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: cli
type: fix
summary: "d8 ssh / d8 scp now use the local OpenSSH/SCP clients via ProxyCommand; pass extra client options through a single --ssh-args flag and the command via --."
impact_level: low
```


src/cli/internal/cmd/ssh/ssh.go
```diff
@@ -161,7 +163,7 @@ func (o *SSH) Run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	client, defaultNamespace, _, err := clientconfig.ClientAndNamespaceFromContext(cmd.Context())
+	_, defaultNamespace, _, err := clientconfig.ClientAndNamespaceFromContext(cmd.Context())
 	if err != nil {
 		return err
 	}
```
diafour, line 170: As client is not needed for "local ssh", we can report the error and proceed with "default" as namespace.

yaroslavborbat (approved):

PR #2477, opened 2026-06-10
chore(ci): allow switching image source on nested e2e clusters

universal-itengineer:
## Description

Make the HTTP image source for the nightly e2e workflows configurable through a single base URL, and default it to the in-cluster image share.

A single optional input `e2e_image_base_url` is added to the reusable nightly pipeline and mapped to the `E2E_IMAGE_BASE_URL` environment variable used by the e2e tests:

* empty value → HTTP images are downloaded from the default external S3 storage (previous behavior);
* non-empty value → it is used as the base URL for HTTP test images.

How it is wired:

* `.github/workflows/e2e-nightly.yml` — a workflow-level `env.E2E_IMAGE_BASE_URL` sets the default source. The `set-vars` job resolves it into the `e2e_image_base_url` output and passes it to both nightly pipelines (replicated and nfs).
* `.github/workflows/e2e-nightly-reusable-pipeline.yml` — declares the optional `e2e_image_base_url` input and exposes it as `E2E_IMAGE_BASE_URL` on the `task run:ci` step.
* `test/e2e/internal/object/precreated_cvi.go` — the precreated CVI URLs are now built via an `imageURL()` helper that reads `E2E_IMAGE_BASE_URL` and falls back to the default S3 base URL when unset. Container image references are left unchanged.
* `test/e2e/scripts/e2e-ci.sh` — `rewrite_testdata_image_urls()` rewrites the HTTP base URL in the legacy `/tmp/testdata` manifests from the default storage to the requested base URL (only when a non-default URL is set and the testdata directory exists).

By default the nightly workflow sets `E2E_IMAGE_BASE_URL` to the in-cluster image share, so both nightly clusters (replicated and nfs) pull HTTP test images from inside the cluster. Clearing the value falls back to the default S3 source.

Nightly nested-cluster e2e runs download HTTP test images from external S3 storage, which has been unreliable. The cluster already hosts a local image share (`e2e-test-images/share`), so nightly runs can use the in-cluster share by default while keeping S3 as an easy fallback.

* Nightly e2e defaults to pulling HTTP test images from the in-cluster image share for both the replicated and nfs clusters.
* Setting `E2E_IMAGE_BASE_URL` to another URL uses that source; clearing it falls back to the default S3 storage.
* Container image references are not changed.

Test info:

* Module: `test/e2e` — `go build ./...` passes, 0 lint issues.
* Full nested-cluster e2e pipeline is being validated via nightly dispatch.

## Checklist

* [ ] The code is covered by unit tests.
* [ ] e2e tests passed.
* [ ] Documentation updated according to the changes.
* [ ] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: ci
type: chore
summary: "Make the nightly e2e HTTP image source configurable via a single base URL and default it to the in-cluster image share."
impact_level: low
```

### DIFF
--- a/.github/workflows/e2e-nightly-reusable-pipeline.yml
+++ b/.github/workflows/e2e-nightly-reusable-pipeline.yml
@@ -111,8 +111,8 @@ on:
       e2e_incluster_image_base_url:
         required: false
         type: string
-        default: "http://share.e2e-test-images.svc.cluster.local"
-        description: "In-cluster base URL for HTTP images used by E2E tests"
+        default: "https://e2e-test-images.e2e.virtlab.flant.com"
+        description: "In-cluster base URL for HTTP images used by E2E tests (image share Ingress reachable from nested clusters)"
       nested_cluster_network_name:
         required: false
         type: string
--- a/.github/workflows/e2e-nightly-reusable-pipeline.yml
+++ b/.github/workflows/e2e-nightly-reusable-pipeline.yml
@@ -574,7 +574,7 @@ jobs:
           FOCUS: ${{ inputs.e2e_focus_tests }}
           E2E_IMAGE_BASE_URL: ${{ inputs.e2e_image_base_url }}
         working-directory: ./test/e2e/
-        run: FOCUS="VirtualMachineConfiguration" task run:ci
+        run: task run:ci
 
       - name: Upload summary test results (json)
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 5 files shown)


danilrwx (dismissed):

universal-itengineer:
https://github.com/deckhouse/virtualization/actions/runs/27365321122/job/80876520213

diafour (approved):

PR #2479, opened 2026-06-11
fix(core): use add op for absent metadata maps in reconciler patch

universal-itengineer:
## Description
The generic reconciler (`reconciler.Resource`) built JSON Patch operations for
`/metadata/finalizers`, `/metadata/labels` and `/metadata/annotations` using the
`replace` op. Under strict RFC 6902 semantics (evanphx json-patch v5, used by the
API server) `replace` is rejected when the target path does not exist, so the patch
was rejected whenever the object had no finalizers/labels/annotations yet.

`JSONPatchOpsForFinalizers`, `JSONPatchOpsForLabels` and `JSONPatchOpsForAnnotations`
now emit an `add` op (which creates the field) when the current value is empty/absent,
and keep `replace` when the field already exists. `add` is accepted under both v4 and
v5 semantics. The `test` op for labels/annotations is preserved to retain the
optimistic lock.

`firmware-update` VMOPs are created by the workload-updater with an annotation but
no labels and no finalizers. On reconcile the controller adds the `vmop-cleanup`
finalizer and the `virtualization.deckhouse.io/virtual-machine-uid` label, producing
a patch like:

```
[{"op":"replace","path":"/metadata/finalizers","value":[...]},
 {"op":"test","path":"/metadata/labels","value":null},
 {"op":"replace","path":"/metadata/labels","value":{...}}]
```

The API server rejected it ("the server rejected our request due to an error in our
request") because `replace` on the missing `/metadata/finalizers` path fails first.
Since the patch is atomic, the finalizer and label were never written, so the VMOP
looped on the same error and never created the `VirtualMachineInstanceMigration`. A
subsequent migration then got stuck (progress reported, but the finalizer/label patch
kept failing). This affects any object created without finalizers/labels/annotations,
not only firmware-update VMOPs.

1. Trigger a firmware update that requires migration on a VM created without labels
   (e.g. `vm-alpine-local-hp-local`).
2. The `firmware-update-*` VMOP receives its finalizer and `virtual-machine-uid`
   label without the "server rejected our request" error.
3. The `VirtualMachineInstanceMigration` is created and the migration proceeds to
   completion.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: core
type: fix
summary: Fix metadata patch failing with "the server rejected our request" for objects without finalizers, labels or annotations.
impact_level: low
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/reconciler/resource.go
+++ b/images/virtualization-artifact/pkg/controller/reconciler/resource.go
@@ -237,16 +237,26 @@ func (r *Resource[T, ST]) JSONPatchOpsForFinalizers() []patch.JSONPatchOperation
 }
 
 func (r *Resource[T, ST]) JSONPatchOpsForAnnotations() []patch.JSONPatchOperation {
-	return []patch.JSONPatchOperation{
-		patch.NewJSONPatchOperation(patch.PatchTestOp, "/metadata/annotations", r.currentObj.GetAnnotations()),
-		patch.NewJSONPatchOperation(patch.PatchReplaceOp, "/metadata/annotations", r.changedObj.GetAnnotations()),
-	}
+	return jsonPatchOpsForMetadataMap("/metadata/annotations", r.currentObj.GetAnnotations(), r.changedObj.GetAnnotations())
 }
 
 func (r *Resource[T, ST]) JSONPatchOpsForLabels() []patch.JSONPatchOperation {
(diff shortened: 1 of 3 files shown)


danilrwx (approved):

deckhouse-BOaTswain:
Cherry pick PR [2482](https://github.com/deckhouse/virtualization/pull/2482) to the branch [release-1.9](https://github.com/deckhouse/virtualization/tree/release-1.9) successful!

PR #2247, opened 2026-04-20
feat(vm): implement CPU/memory hotplug via in-place resize

yaroslavborbat:
## Description

Implement hotplug CPU/Memory via in-place resize (without live migration).

New alpha feature gate added:
- `HotplugCPUAndMemoryWithInPlaceResize` — CPU and memory hotplug using in-place pod resize

KubeVirt patched to expose the in-place resize capability. The `workload-updater` controller updated with a new `hotplug` handler that applies in-place resize when the feature gate is enabled. CPU/memory comparators updated to detect spec changes and trigger reconciliation accordingly.
https://github.com/deckhouse/3p-kubevirt/pull/103

## Why do we need it, and what problem does it solve?

The existing hotplug via live migration requires a second node and causes VM downtime during migration. In-place resize allows changing vCPU count and memory of a running VM without migration, reducing overhead and enabling hotplug in single-node environments.

## What is the expected result?

1. Enable `HotplugCPUAndMemoryWithInPlaceResize` feature gate.
2. Edit `VirtualMachine` spec to change CPU/memory.
3. VM resources are updated without migration or restart.

## Checklist
- [X] The code is covered by unit tests.
- [X] e2e tests passed.
- [ ] Documentation updated according to the changes.
- [X] Changes were tested in the Kubernetes cluster manually.

## Changelog entries

```changes
section: vm
type: feature
summary: Add hotplug CPU and memory via in-place resize (alpha, behind feature gate HotplugCPUAndMemoryWithInPlaceResize).
```

PR #2481, opened 2026-06-11
chore(e2e): Add prechecks for Hotplug resources tests

diafour:
## Description

- Replace explicit Skips with prechecks to enable Hotplug resources tests in compatible clusters.

We want to run hotplug resources tests in compatible clusters.

Hotplug resources tests run in clusters with enabled feature gates.

## Checklist
- [ ] e2e tests passed.


diafour:
Implemented in #2247

PR #2496, opened 2026-06-16
fix(vm): avoid restartAwaitingChanges flicker for hotplug block devices

eofff:
## Description

When `enableParavirtualization` is enabled, VM block devices are handled through hotplug. Because of that, changes in `blockDeviceRefs` must not be reported by the VM change comparator as changes that require restart handling.

This PR stops reporting `blockDeviceRefs` add, remove, and reorder operations when paravirtualization is enabled in both the current and desired VM specs. Unit tests cover these hotplug block device scenarios and verify that they do not produce VM spec changes.

With paravirtualization enabled, attaching or detaching disks is a hotplug operation and should not affect the VM restart-awaiting-changes state.

Before this change, `blockDeviceRefs` updates were still detected as VM spec changes. As a result, `restartAwaitingChanges` could briefly appear or flicker while disks were being connected, disconnected, or reordered, even though these operations do not require a VM restart in the paravirtualized flow.

For a VM with `enableParavirtualization: true` in both the current and desired specs:

- Adding a block device does not trigger `restartAwaitingChanges`.
- Removing a block device does not trigger `restartAwaitingChanges`.
- Reordering block devices does not trigger `restartAwaitingChanges`.
- Hotplug disk operations do not make the VM look like it has pending restart-required changes.

For VMs without paravirtualization, the existing restart-required behavior for block device changes is preserved.

## Checklist
- [x] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
+++ b/images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
@@ -159,8 +159,11 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 		changes = h.detectSpecChanges(ctx, kvvm, &current.Spec, lastAppliedSpec)
 		if !changes.IsEmpty() {
 			kvvmi, kvvmiErr := s.KVVMI(ctx)
-			if kvvmiErr == nil && hasNonHotpluggableVolumes(kvvmi) {
-				changes.UpgradeBlockDeviceChangesToRestart()
+			if kvvmiErr == nil {
+				nonHotpluggableVolumes := nonHotpluggableVolumeRefs(kvvmi)
+				changes.UpgradeBlockDeviceChangesToRestartIf(func(change vmchange.FieldChange) bool {
+					return blockDeviceChangeTouchesRefs(change, nonHotpluggableVolumes)
+				})
 			}
 			// Require restart for CPU and memory changes if VM is non migratable.
 			if h.isVMNonMigratable(current) {
@@ -193,7 +196,7 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 	if kvvm == nil || changes.IsEmpty() {
 		changed.Status.RestartAwaitingChanges = nil
 	} else {
-		changed.Status.RestartAwaitingChanges, err = changes.ConvertPendingChanges()
+		changed.Status.RestartAwaitingChanges, err = changes.ConvertPendingRestartChanges()
 		if err != nil {
 			err = fmt.Errorf("failed to generate pending configuration changes: %w", err)
 			cbConfApplied.
@@ -201,6 +204,9 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 				Reason(vmcondition.ReasonConfigurationNotApplied).
 				Message(service.CapitalizeFirstLetter(err.Error()) + ".")
 			return reconcile.Result{}, err
+		}
+		if len(changed.Status.RestartAwaitingChanges) == 0 {
+			changed.Status.RestartAwaitingChanges = nil
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


images/virtualization-artifact/pkg/controller/vmchange/spec_changes.go
```diff
@@ -230,8 +230,22 @@ func (s *SpecChanges) GetRestartMessages() []string {
 }
 
 func (s *SpecChanges) ConvertPendingChanges() ([]apiextensionsv1.JSON, error) {
```
LopatinDmitr: Is this code unused?

images/virtualization-artifact/pkg/controller/vmchange/spec_changes.go
```diff
@@ -243,9 +257,13 @@ func (s *SpecChanges) ConvertPendingChanges() ([]apiextensionsv1.JSON, error) {
 }
 
 func (s *SpecChanges) UpgradeBlockDeviceChangesToRestart() {
+	s.UpgradeBlockDeviceChangesToRestartIf(nil)
+}
+
+func (s *SpecChanges) UpgradeBlockDeviceChangesToRestartIf(shouldUpgrade func(FieldChange) bool) {
```
LopatinDmitr: May be UpgradeBlockDeviceChangesToRestartWhere? "...If" doesn't look good

images/virtualization-artifact/pkg/controller/vmchange/spec_changes.go
```diff
@@ -243,9 +257,13 @@ func (s *SpecChanges) ConvertPendingChanges() ([]apiextensionsv1.JSON, error) {
 }
 
 func (s *SpecChanges) UpgradeBlockDeviceChangesToRestart() {
```
LopatinDmitr: It's only used in tests. Can we remove it?

images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm.go
```diff
@@ -202,6 +205,9 @@ func (h *SyncKvvmHandler) Handle(ctx context.Context, s state.VirtualMachineStat
 				Message(service.CapitalizeFirstLetter(err.Error()) + ".")
 			return reconcile.Result{}, err
 		}
+		if len(changed.Status.RestartAwaitingChanges) == 0 {
+			changed.Status.RestartAwaitingChanges = nil
```
LopatinDmitr: I don't understand why these lines are

images/virtualization-artifact/pkg/controller/vm/internal/sync_kvvm_test.go
```diff
@@ -578,6 +628,46 @@ var _ = Describe("SyncKvvmHandler", func() {
...
+					PersistentVolumeClaim: &virtv1.PersistentVolumeClaimVolumeSource{Hotpluggable: true},
+				},
+			},
+		}
+
+		nonHotpluggableRefs := nonHotpluggableVolumeRefs(kvvmi)
+		changes := vmchange.SpecChanges{}
+		changes.Add(
+			vmchange.FieldChange{Path: "blockDeviceRefs.0", Operation: vmchange.ChangeRemove, CurrentValue: dataRef, ActionRequired: vmchange.ActionApplyImmediate},
+			vmchange.FieldChange{Path: "blockDeviceRefs.1", Operation: vmchange.ChangeAdd, DesiredValue: extraRef, ActionRequired: vmchange.ActionApplyImmediate},
```
LopatinDmitr, line 656: This DesiredValue (extraRef) is intentionally NOT in the non-hotpluggable set — it only exercises the negative path. There's no positive case where DesiredValue points to a non-hotpluggable volume and the change is expected to upgrade to ActionRestart. Consider adding a ChangeAdd/ChangeReplace change with DesiredValue: rootRef expecting ActionRestart, to cover the DesiredValue branch in blockDeviceChangeTouchesRefs.

LopatinDmitr (approved):

PR #2506, opened 2026-06-17
fix(vm): revert migrated volumes when kvvmi is missing

yaroslavborbat:
## Description
The fix restores migrated volumes in `kvvm.spec` when `kvvmi` is already missing.

It also fixes the VMBDA case: existing hotplug volumes are now rebuilt from the current `VirtualDisk.Status.Target.PersistentVolumeClaim`, so after migration rollback they no longer keep a stale target PVC in `kvvm.spec`.

A user could hit this during the following flow:
- a VM uses at least one local-storage disk, including a disk attached through VMBDA;
- VM eviction starts volume migration;
- migration fails;
- the user uncordons nodes and requests VM restart;
- `kvvmi` is already deleted, while `kvvm.spec` still points to migrated target PVCs.

In that state KubeVirt sets `ManualRecoveryRequired` and refuses to start the VM. The controller could not revert volumes without `kvvmi`, and VMBDA hotplug volumes could also keep stale PVCs because they were preserved in `kvvm.spec` but not rebuilt from the current disk state.

This change makes the controller restore source PVCs in both cases, which unblocks VM recovery after failed migration.

Reproduce with a VM that has a local-storage disk, including a VMBDA-attached disk:
- cordon all nodes;
- start VM eviction;
- wait until the eviction/migration operation fails;
- uncordon nodes;
- request VM restart.

Expected result:
- migrated target PVCs are restored back to source PVCs in `kvvm.spec`;
- VMBDA hotplug volumes are also rebuilt from the current PVC state;
- `ManualRecoveryRequired` no longer blocks recovery in this case;
- the VM starts again without manual `kvvm.spec` editing.

## Checklist
- [x] e2e tests passed.

## Changelog entries
```changes
section: vm
type: fix
summary: "A VM can start again after a failed local-disk migration restart when `kvvmi` is already deleted and migrated volumes, including VMBDA-attached volumes, must be restored from target PVCs back to the current source PVCs."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
+++ b/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
@@ -355,16 +355,19 @@ func setVMBDABlockDeviceDisk(
 ) error {
 	switch ref.Kind {
 	case v1alpha2.VMBDAObjectRefKindVirtualDisk:
+		name := GenerateVDDiskName(ref.Name)
 		vd, ok := vdByName[ref.Name]
 		if !ok || vd == nil {
-			return fmt.Errorf("unexpected error: virtual disk %q should exist in the cluster; please recreate it", ref.Name)
-		}
-
-		if vd.Status.Phase == v1alpha2.DiskTerminating || vd.Status.Target.PersistentVolumeClaim == "" {
+			removeDisk(kvvm, name)
 			return nil
 		}
 
-		return kvvm.SetDisk(GenerateVDDiskName(ref.Name), SetDiskOptions{
+		if vd.Status.Phase == v1alpha2.DiskTerminating || vd.Status.Target.PersistentVolumeClaim == "" {
+			removeDisk(kvvm, name)
+			return nil
+		}
+
+		return kvvm.SetDisk(name, SetDiskOptions{
 			PersistentVolumeClaim: ptr.To(vd.Status.Target.PersistentVolumeClaim),
 			Serial:                GenerateSerialFromObject(vd),
 			IsHotplugged:          true,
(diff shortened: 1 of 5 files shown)


images/virtualization-artifact/pkg/controller/kvbuilder/kvvm_utils.go
```diff
@@ -315,6 +346,38 @@ func setBlockDeviceDisk(
...
+	ref v1alpha2.VMBDAObjectRef,
+	vdByName map[string]*v1alpha2.VirtualDisk,
+	viByName map[string]*v1alpha2.VirtualImage,
+	cviByName map[string]*v1alpha2.ClusterVirtualImage,
+) error {
+	switch ref.Kind {
+	case v1alpha2.VMBDAObjectRefKindVirtualDisk:
+		vd, ok := vdByName[ref.Name]
+		if !ok || vd == nil {
+			return fmt.Errorf("unexpected error: virtual disk %q should exist in the cluster; please recreate it", ref.Name)
```
loktev-d: probably should also check for VD terminating phase to avoid this error

images/virtualization-artifact/pkg/controller/vm/internal/service/migration_volumes.go
```diff
@@ -104,8 +104,18 @@ func (s MigrationVolumesService) SyncVolumes(ctx context.Context, vmState state.
...
-		log.Info("Virtualmachine or kvvmi is nil, skip volume migration.")
+	if kvvmInCluster == nil {
+		log.Info("Virtualmachine is nil, skip volume migration.")
+		return reconcile.Result{}, nil
+	}
+
+	if kvvmiInCluster == nil {
+		if s.shouldPatchVolumes(kvvmInCluster, builtKVVM) {
+			log.Info("kvvmi is nil, force revert kvvm volumes to source volumes.")
+			return reconcile.Result{}, s.patchVolumes(ctx, builtKVVM)
```
loktev-d, line 115: does this also reset UpdateVolumeStrategy?

danilrwx: yes, it does

loktev-d (approved):

PR #2523, opened 2026-06-23
chore(ci): drop svace builder images from image ternaries

nevermarine:
## Description
Replace `builder/golang-alt-svace-*` builder images with `builder/golang-alt-1.25`
in the `SVACE_ENABLED == "false"` branch of `fromImage` ternaries across 16
`werf.inc.yaml` files. The svace-specific builder images are no longer used; the
standard `builder/golang-alt-1.25` image is used instead for both branches.

The `builder/golang-alt-svace-1.24` / `builder/golang-alt-svace-1.25` images are
being dropped from the build pipeline. The svace-enabled builds now use the same
`builder/golang-alt-1.25` base as regular builds, so the svace image variants are
no longer needed. This keeps the ternaries referencing an image that actually
exists and builds successfully.

1. `SVACE_ENABLED` builds use `builder/golang-alt-1.25` instead of the removed
   `builder/golang-alt-svace-*` images.
2. Non-svace builds are unaffected (same images as before).
3. All affected images build successfully in CI.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/bounder/werf.inc.yaml
+++ b/images/bounder/werf.inc.yaml
@@ -12,7 +12,7 @@ imageSpec:
 ---
 image: {{ .ModuleNamePrefix }}{{ .ImageName }}-cbuilder
 final: false
-fromImage: builder/golang-alt-1.24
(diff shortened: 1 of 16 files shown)


nevermarine (commented):
-

universal-itengineer (approved):

PR #2524, opened 2026-06-23
feat(vm): allow disabling automatic firmware update migrations

eofff:
## Description
Added a ModuleConfig annotation-driven switch for automatic firmware update migrations.

When `virtualization.deckhouse.io/disable-firmware-update=true` is set on `ModuleConfig/virtualization`, the existing module hook copies this value into internal Helm values, the controller Deployment receives `DISABLE_FIRMWARE_UPDATE=true`, and the workload-updater firmware handler skips creating `firmware-update-*` VMOPs.

The controller no longer needs to read `ModuleConfig` at runtime for this check.

Operators need a way to temporarily prevent system-initiated firmware update migrations while keeping the module and VM status reconciliation running.

Using the ModuleConfig annotation keeps the switch in the module configuration flow and avoids adding runtime `ModuleConfig` reads or extra controller RBAC permissions.

Set the annotation:

```bash
d8 k annotate moduleconfig virtualization virtualization.deckhouse.io/disable-firmware-update=true --overwrite
```

After the module values are rendered and the controller Deployment is updated, VMs may still report `FirmwareUpToDate=False`, but workload-updater does not create `firmware-update-*` VMOPs while the switch is enabled.

Remove or set the annotation to `false` to restore automatic firmware update migrations.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/virtualization-artifact/pkg/controller/workload-updater/internal/handler/firmware.go
+++ b/images/virtualization-artifact/pkg/controller/workload-updater/internal/handler/firmware.go
@@ -40,22 +40,24 @@ type firmwareMigration interface {
 	OnceMigrate(ctx context.Context, vm *v1alpha2.VirtualMachine, annotationKey, annotationExpectedValue string) (bool, error)
 }
 
-func NewFirmwareHandler(client client.Client, migration firmwareMigration, firmwareImage, namespace, virtControllerName string) *FirmwareHandler {
+func NewFirmwareHandler(client client.Client, migration firmwareMigration, firmwareImage, namespace, virtControllerName string, disableFirmwareUpdate bool) *FirmwareHandler {
 	return &FirmwareHandler{
-		client:             client,
-		oneShotMigration:   migration,
-		firmwareImage:      firmwareImage,
-		namespace:          namespace,
-		virtControllerName: virtControllerName,
+		client:                client,
+		oneShotMigration:      migration,
+		firmwareImage:         firmwareImage,
+		namespace:             namespace,
+		virtControllerName:    virtControllerName,
+		disableFirmwareUpdate: disableFirmwareUpdate,
 	}
 }
 
 type FirmwareHandler struct {
-	client             client.Client
-	oneShotMigration   firmwareMigration
-	firmwareImage      string
-	namespace          string
-	virtControllerName string
+	client                client.Client
+	oneShotMigration      firmwareMigration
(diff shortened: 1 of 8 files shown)


images/virtualization-artifact/pkg/controller/workload-updater/internal/handler/firmware.go
```diff
@@ -96,6 +110,22 @@ func (h *FirmwareHandler) needUpdate(vm *v1alpha2.VirtualMachine) bool {
 	return false
 }
 
+func (h *FirmwareHandler) firmwareUpdateDisabled(ctx context.Context) (bool, error) {
+	if h.client == nil {
```
Isteb4k: Can it be nil?

images/virtualization-artifact/pkg/controller/workload-updater/internal/handler/firmware.go
```diff
@@ -96,6 +110,22 @@ func (h *FirmwareHandler) needUpdate(vm *v1alpha2.VirtualMachine) bool {
...
+		return false, nil
+	}
+
+	moduleConfig := &mcapi.ModuleConfig{}
+	if err := h.client.Get(ctx, client.ObjectKey{Name: virtualizationModuleName}, moduleConfig); err != nil {
+		if apierrors.IsNotFound(err) {
+			return false, nil
```
Isteb4k: err

images/virtualization-artifact/pkg/controller/workload-updater/internal/handler/firmware.go
```diff
@@ -96,6 +110,22 @@ func (h *FirmwareHandler) needUpdate(vm *v1alpha2.VirtualMachine) bool {
...
 }
 
+func (h *FirmwareHandler) firmwareUpdateDisabled(ctx context.Context) (bool, error) {
+	if h.client == nil {
+		return false, nil
+	}
+
+	moduleConfig := &mcapi.ModuleConfig{}
+	if err := h.client.Get(ctx, client.ObjectKey{Name: virtualizationModuleName}, moduleConfig); err != nil {
+		if apierrors.IsNotFound(err) {
```
eofff: remove this

Isteb4k (changes requested):

Isteb4k (approved):

PR #2527, opened 2026-06-23
fix(uploader): sync ingress with publicDomainTemplate via server-side apply

fl64:
## Description

Changing the cluster domain (`publicDomainTemplate`) left existing image-upload endpoints on the old domain, so the upload URL shown for an already-created `VirtualImage` / `VirtualDisk` / `ClusterVirtualImage` (with `dataSource.type=Upload`) kept pointing at the old host and uploads to it failed.

The upload endpoint was created once when an image or disk with `Upload` source first appeared and was never reconciled afterwards. After an administrator changed the cluster domain, the controller restarted with the new domain, but the already-existing endpoints kept their original host. Users kept seeing the old upload URL, so uploading data to an already-created image broke until the image was deleted and recreated.

The endpoint is now reconciled on every upload-flow pass, so its host, TLS host, and upload URL follow the current cluster domain automatically.

1. Create a `VirtualImage` with `dataSource.type=Upload` — the upload URL matches the current cluster domain.
2. Change `publicDomainTemplate` and wait for the controller to pick up the new domain.
3. The upload URL and endpoint host update to the new domain; uploading to the existing image keeps working.

## Checklist

- [ ] e2e tests passed.

## Changelog entries

```changes
section: images
type: fix
summary: "Upload endpoint now follows publicDomainTemplate changes instead of keeping the host it was created with."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
+++ b/images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
@@ -52,12 +52,12 @@ func NewIngress(settings *IngressSettings) *Ingress {
 	return &Ingress{settings}
 }
 
+const fieldOwner = "virtualization-controller"
+
 // Apply reconciles the Ingress using server-side apply: it creates the Ingress
 // if absent and updates spec/annotations when they drift (e.g. when
 // publicDomainTemplate changes UPLOADER_INGRESS_HOST). Field ownership is
 // stable across controller restarts.
-const fieldOwner = "virtualization-controller"
-
 func (i *Ingress) Apply(ctx context.Context, c client.Client) (*netv1.Ingress, error) {
 	desired := i.makeSpec()
 	desired.SetGroupVersionKind(netv1.SchemeGroupVersion.WithKind("Ingress"))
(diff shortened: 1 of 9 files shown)


images/virtualization-artifact/pkg/controller/uploader/uploader_ingress.go
```diff
@@ -52,14 +52,21 @@ func NewIngress(settings *IngressSettings) *Ingress {
...
-	ing := i.makeSpec()
+// Apply reconciles the Ingress using server-side apply: it creates the Ingress
+// if absent and updates spec/annotations when they drift (e.g. when
+// publicDomainTemplate changes UPLOADER_INGRESS_HOST). Field ownership is
+// stable across controller restarts.
+const fieldOwner = "virtualization-controller"
 
-	if err := client.Create(ctx, ing); err != nil {
+func (i *Ingress) Apply(ctx context.Context, c client.Client) (*netv1.Ingress, error) {
+	desired := i.makeSpec()
```
LopatinDmitr, line 62: More correctly:
const fieldOwner = "virtualization-controller"

// Apply reconciles the Ingress using server-side apply: ...
func (i *Ingress) Apply(ctx context.Context, c client.Client) (*netv1.Ingress, error) {

fl64: Good catch — the doc comment was attached to the const instead of the function. Moved `const fieldOwner` above the comment so it now documents `Apply`. Fixed.

LopatinDmitr (approved):

PR #2551, opened 2026-06-29
fix(observability): virtualization-audit tls bad record MAC on cert rotation

danilrwx:
## Description

Fix the recurring `D8LogShipperDestinationErrors` alert and the `tls: bad record MAC` log spam on `virtualization-audit`, and harden the audit ingestion path.

Rolling the audit pod on every certificate rotation is a module policy. Rather than avoid the roll, this PR makes the roll clean so it stops dropping audit events.

`server.go`:
- Drain active connections gracefully on shutdown: close them so the `tls.Conn` sends `close_notify` (clean EOF for the peer instead of a truncated record), and wait up to a timeout for in-flight handlers. This is what keeps the policy-driven roll from surfacing as `bad record MAC` / dropped events.
- Enforce mTLS (`RequireAndVerifyClientCert`). Previously `RootCAs` was set but `ClientAuth` was not, leaving the audit ingestion endpoint unauthenticated and the CA loading effectively dead code. The log-shipper already presents a client certificate signed by the same CA.
- Set `MinVersion` to TLS 1.2.
- Raise the `bufio.Scanner` buffer to 4 MiB; the 64 KiB default silently dropped the rest of a connection on large audit events (k8s objects with `managedFields`).
- Log mid-stream connection drops at debug instead of error with a stacktrace — a peer disconnecting is expected churn, not a server fault.

`deployment.yaml`:
- Add TCP readiness and liveness probes on the audit port (there were none), so the Service does not route to a not-yet-ready pod and a hung server is restarted.

`log-config.yaml`:
- Switch the `to-virtualization-audit` destination buffer from Memory to Disk so undelivered audit events survive a transient sink outage and a log-shipper agent restart.

The cluster fired `D8LogShipperDestinationErrors`, and the audit pod logged `local error: tls: bad record MAC`, while the log-shipper logged `certificate verify failed`, `connection reset/refused`, `unexpected EOF` and dropped audit events.

Root cause: the audit pod is rolled on every certificate rotation (the `checksum/secret` annotation, by module policy), and the server tore down the log-shipper's persistent TLS connection abruptly. In the teardown window:
- events in the log-shipper buffer were dropped (the alert),
- a TLS record cut mid-stream surfaced on the server as `bad record MAC` (reproduced deterministically),
- a CA skew between the log-shipper and the freshly-rolled pod surfaced as `certificate verify failed`.

Graceful drain makes the roll clean (the server sends `close_notify`, so the log-shipper sees a clean EOF), the disk buffer lets the log-shipper re-deliver buffered events instead of dropping them, and the probes keep the Service from routing to a not-ready pod. The mTLS and scanner-buffer fixes close an unauthenticated-ingestion hole and a silent audit-event-loss bug found along the way.

1. Rotate the `virtualization-audit-tls` secret (or wait for rotation). The audit pod rolls, as per policy.
2. During the roll the server closes connections cleanly (`close_notify`) and the log-shipper re-delivers buffered events.
3. No `D8LogShipperDestinationErrors`, no `bad record MAC` / `certificate verify failed` around the rotation.
4. Audit events keep flowing; large events (>64 KiB) are no longer truncated.
5. Connections without a valid client certificate are rejected at handshake (mTLS).

## Checklist
- [ ] e2e tests passed.

### Manual verification

The original `local error: tls: bad record MAC` was reproduced deterministically on the cluster (real TLS handshake, then a forged application record written past the TLS layer). On the build with this fix, the same scenario shows:

- **mTLS enforced** — a client with no certificate gets a TLS `handshake failure` (alert 40); a client presenting the audit certificate handshakes successfully.
- **No error spam** — the forged record produces zero `bad record MAC` / `Scanner error` lines at error level (downgraded to debug); the connection is drained instead of logging a stacktrace.
- **Session really established** — the probe line on the authenticated connection surfaced as `warn: Failed to process audit event: ... parsing JSON`, i.e. the exact code path that previously logged `bad record MAC`.

> Still worth a dedicated stand check: the disk-buffer re-delivery behavior of the log-shipper socket sink across a real pod roll.

## Changelog entries
```changes
section: observability
type: fix
summary: Fixed loss of audit events and false `D8LogShipperDestinationErrors` alerts during certificate rotation of the `virtualization-audit` pod.
```

### DIFF
--- a/images/virtualization-artifact/pkg/audit/server/server.go
+++ b/images/virtualization-artifact/pkg/audit/server/server.go
@@ -63,18 +63,12 @@ func (s *tcpServer) Run(ctx context.Context, opts ...Option) error {
 	var listener net.Listener
 	var err error
 	if o.TLS != nil {
-		// Fail fast if the certificate material is missing or invalid at startup.
-		if _, err = loadServerTLSConfig(o.TLS); err != nil {
+		// The pod is rolled on every certificate rotation (module policy), so the
+		// cert is loaded once at startup; graceful shutdown drains connections so
+		// that roll stays clean for the log-shipper.
+		var cfg *tls.Config
+		if cfg, err = loadServerTLSConfig(o.TLS); err != nil {
 			return err
-		}
-
-		// ponytail: reload cert and client CA from disk on every handshake so
-		// secret rotation needs no pod restart. Handshakes are rare (vector keeps
-		// a persistent connection), so per-handshake disk reads are negligible.
-		cfg := &tls.Config{
-			GetConfigForClient: func(*tls.ClientHelloInfo) (*tls.Config, error) {
-				return loadServerTLSConfig(o.TLS)
-			},
 		}
 		listener, err = tls.Listen("tcp", s.addr, cfg)
 	} else {
--- a/images/virtualization-artifact/pkg/audit/server/server.go
+++ b/images/virtualization-artifact/pkg/audit/server/server.go
@@ -25,10 +25,14 @@ import (
 	"log/slog"
 	"net"
 	"os"
-	"time"
 
 	"github.com/deckhouse/deckhouse/pkg/log"
 )
+
+// maxAuditEventSize bounds a single audit event line. The bufio.Scanner default
+// is 64 KiB, which k8s audit events with large objects (managedFields, VM specs)
+// can exceed — dropping the rest of the connection. 4 MiB leaves generous headroom.
+const maxAuditEventSize = 4 * 1024 * 1024
 
 type Server interface {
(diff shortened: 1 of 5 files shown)


universal-itengineer (approved):

deckhouse-BOaTswain:
Backport failed. See [Job](https://github.com/deckhouse/virtualization/actions/runs/28381712360) for details.

deckhouse-BOaTswain:
Cherry pick PR [2555](https://github.com/deckhouse/virtualization/pull/2555) to the branch [release-1.9](https://github.com/deckhouse/virtualization/tree/release-1.9) successful!

PR #2558, opened 2026-06-30
fix(api): make user-facing condition messages clearer

fl64:
## Description

Status conditions on virtualization resources sometimes described a problem in terms the user cannot see or act on — launcher and migration pods, storage volume claims, internal virtualization objects, image-registry internals — or used awkward, sometimes ungrammatical wording. This reworks the `.status.conditions[].message` texts across VirtualMachine, VirtualMachineBlockDeviceAttachment, VirtualDisk, VirtualImage, ClusterVirtualImage, VirtualDiskSnapshot and VirtualMachineSnapshot so each message describes the current state in terms of the resources a user actually works with.

Only the message wording changes. Condition types, reasons, statuses and all reconcile behaviour stay the same. This PR covers the text-only changes; messages that still relay a raw internal error will be cleaned up separately.

A user reads a resource's status to understand what is happening with it. When a VirtualMachine is stuck or a disk or image is not ready, a message that points at a pod, a storage volume claim or an internal object gives no actionable signal and exposes implementation details the user has no control over. Clear, self-explanatory messages let users troubleshoot on their own and read straight into the module's release notes.

`kubectl describe` on these resources shows user-facing wording, for example:
- `Virtual disk "vd1" is waiting for its storage to be provisioned` (was: `… underlying PVC to be bound`);
- `Migration is in progress: preparing the migration target` (was: `… target pod is being scheduled and prepared`);
- `VirtualMachine "x" is not fully started yet` (was: `InternalVirtualizationVirtualMachineInstance "x" not found`);
- `The image is being imported` (was: `Import is in the process of provisioning to DVCR`).

Behaviour is unchanged: reasons and statuses are the same, so anything that keys on them is unaffected.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: VirtualMachine status messages are clearer and no longer expose internal implementation details.
---
section: vmbda
type: fix
summary: VirtualMachineBlockDeviceAttachment status messages are clearer and no longer expose internal implementation details.
---
section: vd
type: fix
summary: VirtualDisk status messages are clearer and no longer expose internal implementation details.
---
section: vi
type: fix
summary: VirtualImage status messages are clearer and no longer expose internal implementation details.
---
section: cvi
type: fix
summary: ClusterVirtualImage status messages are clearer and no longer expose internal implementation details.
---
section: vdsnapshot
type: fix
summary: VirtualDiskSnapshot status messages have clearer wording.
---
section: vmsnapshot
type: fix
summary: VirtualMachineSnapshot status messages have clearer wording.
```


images/virtualization-artifact/pkg/controller/vd/internal/source/sources.go
```diff
@@ -111,14 +111,14 @@ func setPhaseConditionForFinishedDisk(
 	switch {
 	case pvc == nil:
 		newPhase = v1alpha2.DiskLost
-		setReadyConditionWithWFFCAccounting(vd, cb, metav1.ConditionFalse, vdcondition.Lost, fmt.Sprintf("PVC %s not found.", supgen.PersistentVolumeClaim().String()))
+		setReadyConditionWithWFFCAccounting(vd, cb, metav1.ConditionFalse, vdcondition.Lost, fmt.Sprintf("The underlying PersistentVolumeClaim %q was not found.", supgen.PersistentVolumeClaim().String()))
 	case pvc.Status.Phase == corev1.ClaimLost:
 		if pvc.GetAnnotations()[annotations.AnnDataExportRequest] == "true" {
 			newPhase = v1alpha2.DiskExporting
 			setReadyConditionWithWFFCAccounting(vd, cb, metav1.ConditionFalse, vdcondition.Exporting, "PV is being exported")
```
prismagod:
```suggestion
			setReadyConditionWithWFFCAccounting(vd, cb, metav1.ConditionFalse, vdcondition.Exporting, "The PersistentVolume is being exported.")
```

images/virtualization-artifact/pkg/controller/vi/internal/source/step/wait_for_pod_step.go
prismagod:
```suggestion
					Message("Waiting for the PersistentVolumeClaim %q to be Bound.")
```

images/virtualization-artifact/pkg/controller/vdsnapshot/internal/life_cycle.go
```diff
@@ -392,7 +392,7 @@ func (h LifeCycleHandler) Handle(ctx context.Context, vdSnapshot *v1alpha2.Virtu
 		cb.
 			Status(metav1.ConditionFalse).
 			Reason(vdscondition.Snapshotting).
-			Message(fmt.Sprintf("Waiting fot the volume snapshot %q to be ready to use.", vdSnapshot.Name))
+			Message(fmt.Sprintf("Waiting for the snapshot of virtual disk %q to be ready to use.", vdSnapshot.Name))
```
prismagod:
```suggestion
			Message(fmt.Sprintf("Waiting for the snapshot of virtual disk %q to be ready to use.", vdSnapshot.Spec.VirtualDiskName))
```

images/virtualization-artifact/pkg/controller/vd/internal/source/sources.go
fl64: Done, thanks — applied.

images/virtualization-artifact/pkg/controller/vdsnapshot/internal/life_cycle.go
fl64: Good catch, accepted — it was interpolating the snapshot name instead of the disk name. Now uses spec.virtualDiskName.

images/virtualization-artifact/pkg/controller/vi/internal/source/step/wait_for_pod_step.go
fl64: Accepted the wording. Applied without the %q name on purpose: on this branch the pvc may be nil (the unit test passes nil), so a %q here would render empty / risk a nil deref. Message is now "Waiting for the PersistentVolumeClaim to be Bound."

prismagod (dismissed):

Isteb4k (approved):

PR #2564, opened 2026-07-01
feat(vi/cvi): recover VirtualImage and ClusterVirtualImage from ImageLost when data returns

danilrwx:
## Description
Automatically recover `VirtualImage` and `ClusterVirtualImage` that entered the `ImageLost` phase (image is missing in DVCR) **without re-importing the data**.

The `ImagePresenceHandler` already moved a `Ready` image to `ImageLost` when it disappeared from DVCR. It is now symmetric and also handles the reverse transition:

- while the image is `ImageLost`, the handler keeps rechecking DVCR with a backoff derived from the Ready condition's `LastTransitionTime` (`backoff.Progressive` in `common/backoff`): the interval grows from 15s up to 5m the longer the image stays lost, so a prolonged outage does not hammer the registry. It is stateless and survives controller restarts;
- once the image reappears in DVCR (for example, when the DVCR backing PVC is remounted after a temporary loss), the resource is restored to `Ready` — only the phase and the `Ready` condition are flipped, the rest of the status (`Target`, `Size`, `Format`, …) was never cleared;
- a `VirtualImageLostRecovered` / `ClusterVirtualImageLostRecovered` event is emitted on recovery.

Because recovery reuses the blobs already present in DVCR, no re-download happens and **every source recovers, including `Upload`**, whose data cannot be re-fetched.

The `LifeCycleHandler` early-returns on `ImageLost` as before (no import restart). Handler order (`LifeCycleHandler` → `ImagePresenceHandler`) is unchanged and the `ImagePVCLost` phase is not affected. Healthy `Ready` images are not polled — loss detection stays event-driven as before; polling is added only while a resource is lost.

When the DVCR backing storage (its PVC) is temporarily lost and later comes back, all images served by DVCR are physically intact — only the registry was unavailable for a while. Today VI/CVI correctly move to `ImageLost` during the outage, but then stay stuck there even though the blobs returned, forcing users to delete and recreate the resources by hand.

This is especially painful for a mass DVCR outage affecting dozens of images at once. Re-importing is both unnecessary (the data is already there) and impossible for `Upload` sources. Flipping back to `Ready` when the data reappears restores the images automatically, with no re-download and no user action.

1. Create a VI/CVI backed by DVCR, wait for `Ready`.
2. Make DVCR lose the data (detach/remove the DVCR PVC, or delete the blob), wait for the resource to enter `ImageLost`.
3. Bring the data back (remount the DVCR PVC / restore the blob).
4. Within the recheck interval the resource returns to `Ready` on its own; a `VirtualImageLostRecovered` / `ClusterVirtualImageLostRecovered` event is recorded. This works for every source type, including `Upload`.
5. While the data is still missing, the resource stays in `ImageLost` and DVCR keeps being rechecked with a backoff that grows the longer it stays lost (up to 5m).

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: core
type: feature
summary: Automatically restore VirtualImage and ClusterVirtualImage from the ImageLost phase once the image reappears in DVCR, without re-importing the data.
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/cvi/internal/image_presence.go
+++ b/images/virtualization-artifact/pkg/controller/cvi/internal/image_presence.go
@@ -25,6 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/deckhouse/virtualization-controller/pkg/common/backoff"
 	"github.com/deckhouse/virtualization-controller/pkg/controller/conditions"
 	"github.com/deckhouse/virtualization-controller/pkg/dvcr"
 	"github.com/deckhouse/virtualization-controller/pkg/eventrecord"
@@ -32,9 +33,14 @@ import (
 	"github.com/deckhouse/virtualization/api/core/v1alpha2/cvicondition"
 )
 
-// imageLostRecheckInterval is how often DVCR is polled while an image is lost,
-// so recovery is noticed shortly after the data (for example, a DVCR PVC) returns.
-const imageLostRecheckInterval = 30 * time.Second
+// While an image is lost, DVCR is rechecked with an exponential backoff
+// (starting from imageLostRecheckBase, capped at imageLostRecheckMax) so a
+// prolonged DVCR outage does not keep hammering the registry, yet recovery is
+// still noticed once the data (for example, a DVCR PVC) returns.
+const (
+	imageLostRecheckBase = 15 * time.Second
+	imageLostRecheckMax  = 5 * time.Minute
+)
 
 type ImagePresenceHandler struct {
 	imageChecker dvcr.ImageChecker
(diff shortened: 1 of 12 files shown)


images/virtualization-artifact/pkg/controller/cvi/internal/image_presence.go
```diff
@@ -55,13 +65,37 @@ func (h *ImagePresenceHandler) Handle(ctx context.Context, cvi *v1alpha2.Cluster
...
+				Generation(cvi.Generation).
+				Status(metav1.ConditionFalse).
+				Reason(cvicondition.ImageLost).
+				Message("The image data is no longer available and needs to be recreated.")
+
+			conditions.SetCondition(cb, &cvi.Status.Conditions)
+		}
+
+		// Keep polling: the data may return (for example, when the DVCR PVC is remounted).
+		return reconcile.Result{RequeueAfter: imageLostRecheckInterval}, nil
```
loktev-d: Probably should add backoff or timeout in case the image is gone forever

images/virtualization-artifact/pkg/controller/vi/internal/image_presence.go
```diff
@@ -70,13 +81,37 @@ func (h *ImagePresenceHandler) Handle(ctx context.Context, vi *v1alpha2.VirtualI
...
+				Generation(vi.Generation).
+				Status(metav1.ConditionFalse).
+				Reason(vicondition.ImageLost).
+				Message("The image data is no longer available and needs to be recreated.")
+
+			conditions.SetCondition(cb, &vi.Status.Conditions)
+		}
+
+		// Keep polling: the data may return (for example, when the DVCR PVC is remounted).
+		return reconcile.Result{RequeueAfter: imageLostRecheckInterval}, nil
```
loktev-d: Probably should add backoff or timeout in case the image is gone forever

loktev-d (approved):

PR #2572, opened 2026-07-02
feat(vmpool): add VirtualMachinePool for group VM management

fl64:
## Description

Virtualization has no primitive to manage a **group** of identical virtual machines whose count changes over time. Every "I need N identical VMs and the number varies" scenario — CI runner fleets, VDI desktop pools — has to be solved with orchestration outside the platform: users write their own controllers/scripts to create and delete `VirtualMachine`s, watch their number, recreate lost ones and clean up their disks. This duplicates logic and is error-prone around races and node failures.

This PR introduces **`VirtualMachinePool`** (paid editions only, EE/SE+): a namespaced resource that declaratively keeps a requested number of identical VMs and integrates with `kubectl scale`, HPA and KEDA through the standard `scale` subresource. Its template is an ordinary `VirtualMachineSpec`, so a replica is no different from a manually created VM.

The feature is complete and covers:

- **Replica management** — keeps the requested number of identical VMs, replaces lost ones, and reports state in `status` (`replicas`, `readyReplicas`, `selector`, conditions). It is cache-lag-safe (ReplicaSet-style expectations), so a lagging informer cache cannot double-create replicas.
- **Scale-down policy** — `NewestFirst` / `OldestFirst` choose which replica leaves on anonymous scale-down via `scale`; `Explicit` forbids anonymous shrink through a webhook, so for "busy" workloads (CI runners, VDI) replicas can only be removed by address.
- **Addressed removal** — the `scaleDownWith` subresource deletes named replicas and shrinks the pool by that count, instead of letting the controller pick victims.
- **In-place template propagation** — editing the template rolls the change out to existing replicas; disruptive changes wait for a restart and are surfaced via `status.restartPendingReplicas`.
- **Reusable disks** — per-replica disks are described in `virtualDiskTemplates` with a `reclaim` policy: `Delete` disks belong to the VM and are removed with it; `Retain` disks belong to the pool, outlive the replica and are reused on the next scale-up, with an optional warm buffer (`keep`) and TTL garbage-collection.

The resource is available only in paid editions, gated behind the `VirtualMachinePool` module feature gate (default off, locked off in CE). The API/CRD installs in every edition, but the controller self-gates on the feature gate, so the resource does nothing in CE.

Two mass scenarios suffer most: CI/CD runners (GitLab Runner autoscaling expects a backend that can "give me N more" and reclaim idle ones) and VDI pools (warm desktops that self-heal on node failure). Without a group primitive, DVP cannot serve these natively and each team reinvents the orchestration, usually with bugs in race and failure handling. `VirtualMachinePool` gives users a native, declarative backend for autoscaling fleets of VMs without writing their own replica controller, and lets heavy per-replica disks be reused across scale cycles instead of rebuilt every time.

With the `VirtualMachinePool` feature gate enabled (EE/SE+):

1. Create a pool with `spec.replicas: N`, a `spec.virtualMachineTemplate` and `spec.virtualDiskTemplates` — the controller converges the number of `VirtualMachine`s to `N`, each with its own disks (the first disk template is the boot device).
2. `kubectl scale virtualmachinepool/<name> --replicas=M` (or HPA/KEDA) scales the pool; `scaleDownPolicy` decides which replicas leave, or forbids anonymous shrink for `Explicit`.
3. `scaleDownWith` removes specific replicas and shrinks the pool by that count; a plain `kubectl delete vm` is treated as a lost replica and replaced.
4. A member in `Stopped` is kept, not duplicated.
5. Editing the template propagates to replicas; disruptive edits wait for a restart (`status.restartPendingReplicas`).
6. `Retain` disks freed on scale-down are reused on the next scale-up instead of being recreated; deleting the pool cascades its replicas and pool-owned disks.
7. `.status` reports `replicas` / `readyReplicas` / `updatedReplicas` / `restartPendingReplicas`, the `selector`, and the `Available` / `Progressing` conditions.

## Checklist
- [x] e2e tests passed. 

## Changelog entries

```changes
section: vmpool
type: feature
summary: "Add VirtualMachinePool (EE/SE+) for declarative group management of virtual machines, scalable via the standard scale subresource, HPA and KEDA."
```


images/virtualization-artifact/cmd/virtualization-controller/setup_enterprise.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/deckhouse/virtualization-controller/pkg/controller/vmpool"
+	"github.com/deckhouse/virtualization-controller/pkg/logger"
+)
+
+// setupEnterpriseControllers wires controllers for paid-edition (EE/SE+)
+// features. Like the other edition-gated controllers (e.g. VolumeMigration), the
+// code always compiles; each controller self-gates on its feature gate at setup
+// (the VirtualMachinePool gate is locked off in CE), so this is a no-op there.
+func setupEnterpriseControllers(
```
yaroslavborbat: This function doesn't add any clarity. I'd prefer to remove it. It's not clear why this controller becomes EE or CE because of it, so you have to dig deeper into the code to understand what's happening.

images/virtualization-artifact/pkg/apiserver/registry/vmpool/rest/scaledownwith.go
```diff
@@ -0,0 +1,150 @@
...
+// done on the main resource (server-side, from the apiserver's own identity),
+// so it bypasses the /scale guard — that is what makes addressed removal work
+// for Explicit pools.
+type ScaleDownWithREST struct {
+	client client.Client
+}
+
+var (
+	_ rest.Storage   = &ScaleDownWithREST{}
+	_ rest.Connecter = &ScaleDownWithREST{}
```
yaroslavborbat: We use the Connecter interface in the VM because we are proxying requests and streams to another API server. There is no need for you to do this. Please consider the Creater interface instead

images/virtualization-artifact/pkg/apiserver/api/install.go
```diff
@@ -68,6 +69,9 @@ func Build(store *storage.VirtualMachineStorage) genericapiserver.APIGroupInfo {
 		"virtualmachines/addresourceclaim":    store.AddResourceClaimREST(),
 		"virtualmachines/removeresourceclaim": store.RemoveResourceClaimREST(),
 	}
+	// Enterprise-only resources (e.g. virtualmachinepools/scaledownwith) are added
+	// by the EE build; the CE build leaves the map untouched.
+	installEnterpriseResources(resourcesV1alpha2, client)
```
yaroslavborbat: Should we install the resources here only if the featureGate is enabled?

images/virtualization-artifact/pkg/apiserver/api/install.go
```diff
@@ -77,12 +81,13 @@ func Install(
...
 	proxyCertManager certmanager.CertificateManager,
+	client client.Client,
 ) error {
 	vmStorage := storage.NewStorage(
 		vmLister,
 		kubevirt,
 		proxyCertManager,
 	)
-	info := Build(vmStorage)
+	info := Build(vmStorage, client)
```
yaroslavborbat: It doesn't look quite right. We created vmStorage and passed it in, but for vmPoolStorage we only passed the dependencies needed to create it

images/virtualization-artifact/pkg/apiserver/registry/vmpool/storage/storage.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// GetSingularName implements rest.SingularNameProvider.
+func (store VirtualMachinePoolStorage) GetSingularName() string {
+	return "virtualmachinepool"
+}
+
+// Get implements rest.Getter. The meta-object is intentionally not served — the
+// client must address a subresource (see package doc / ADR).
+func (store VirtualMachinePoolStorage) Get(_ context.Context, name string, _ *metav1.GetOptions) (runtime.Object, error) {
```
yaroslavborbat: Let's not return NotFound. Instead, let's return it for the resource, similar to how we do it for vm.

images/virtualization-artifact/pkg/apiserver/server/config.go
```diff
@@ -80,10 +83,23 @@ func (c Config) Complete() (*Server, error) {
 		return nil, err
 	}
 
+	// Write-capable client used by enterprise subresources (e.g. scaleDownWith)
+	// to delete pool members and adjust spec.replicas from the apiserver's own
+	// identity.
+	crScheme := runtime.NewScheme()
```
yaroslavborbat: I would prefer using a typed generated client over the controller-runtime client. Also, for VMs we create a separate cache. Is it needed for the pool as well?

images/virtualization-artifact/pkg/controller/vmpool/vmpool_scale_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	mgr.GetWebhookServer().Register(ScaleWebhookPath, &webhook.Admission{
+		Handler: &scaleValidator{client: mgr.GetClient()},
+	})
+}
+
+type scaleValidator struct {
+	client client.Client
+}
+
+func (v *scaleValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
```
yaroslavborbat, line 60: We prefer to describe Create, Update, and Delete operations separately. Take a look at how the webhooks are implemented in other packages

yaroslavborbat (changes requested):

images/virtualization-artifact/cmd/virtualization-controller/setup_enterprise.go
fl64: Done — removed the wrapper; the pool controller and scale webhook are now set up inline in `main.go` like every other controller.

images/virtualization-artifact/pkg/apiserver/registry/vmpool/rest/scaledownwith.go
fl64: Done in 558b4b2 — switched to `rest.NamedCreater`. No proxying here, so no need for `Connecter`.

images/virtualization-artifact/pkg/apiserver/api/install.go
fl64: Done — served only in paid editions. The runtime `VirtualMachinePool` gate isn't wired into the apiserver process, so I gate by the compiled-in edition; the CRD and controller still enforce the runtime gate.

images/virtualization-artifact/pkg/apiserver/api/install.go
fl64: Done — pool storage is now constructed in `Install` and injected, the same way `vmStorage` is.

images/virtualization-artifact/pkg/apiserver/registry/vmpool/storage/storage.go
fl64: Done — returns `NewNotFound` scoped to the resource, same as `vm`.

images/virtualization-artifact/pkg/apiserver/server/config.go
fl64: Done — now uses the generated typed clientset. No separate cache for the pool: reads go straight to the API, since `scaleDownWith` is a rare validate-before-delete mutation where a cache would only risk stale membership.

images/virtualization-artifact/pkg/controller/vmpool/vmpool_scale_webhook.go
fl64: Operations are now dispatched explicitly and the update path is split out. It stays a raw admission handler (not a `CustomValidator`) because it guards the `scale` subresource, whose request object is an `autoscalingv1.Scale` — the builder's typed validator decodes the parent `VirtualMachinePool` and can't target a subresource path.

images/virtualization-artifact/pkg/apiserver/api/install.go
fl64: Correction after cluster testing: gating this by edition was wrong. The `virtualization-api` binary is built without `-tags MODULE_EDITION` and doesn't wire in the feature gate, so inside the apiserver `version.GetEdition()` is always CE and the gate reads compile defaults — the edition check silently dropped the subresource in every build (endpoint 404'd in the cluster). Fixed in a3129e41: register unconditionally; availability is enforced upstream by the CRD install and the self-gating controller. Verified in-cluster (endpoint back, verb `create`) + e2e green.

crds/doc-ru-virtualmachinepools.yaml
```diff
@@ -0,0 +1,97 @@
...
+                          description: |
+                            Метки, добавляемые каждой реплике.
+                        annotations:
+                          description: |
+                            Аннотации, добавляемые каждой реплике.
+                    spec:
+                      description: |
+                        Спецификация виртуальной машины, из которой создаётся каждая реплика (обычный VirtualMachineSpec, см. [VirtualMachine](cr.html#virtualmachine)).
+
+                        Per-replica диски указываются в `blockDeviceRefs` по имени своего шаблона из `virtualDiskTemplates` — так задаётся полный список устройств и порядок загрузки, как в обычной виртуальной машине.
```
prismagod: Per USER_GUIDE.md: the pool template has no `blockDeviceRefs`: disks go in `virtualDiskTemplates`, and the controller handles the rest. This line says the opposite; I'd drop it. Also the English CRD doesn't have it.

images/virtualization-artifact/pkg/apiserver/registry/vmpool/storage/storage.go
yaroslavborbat: See https://github.com/deckhouse/virtualization/blob/3fccd70c9791e69c728ef72b59a50c5bdb7c34f4/images/virtualization-artifact/pkg/apiserver/registry/vm/storage/storage.go#L142

In VM, we return a fake resource for GET requests and only construct a NotFound response if the resource truly doesn't exist. VMPool currently uses different logic. We should make this behavior consistent by updating either VM or VMPool

fl64: Done in 375a56d. `Get` now mirrors the VM storage: it looks the pool up via the typed client and returns a meta-object when it exists, `NewNotFound` only when it truly does not (and `NewInternalError` otherwise). No divergence between VM and VMPool anymore.

crds/doc-ru-virtualmachinepools.yaml
fl64: Good catch, confirmed against the code (`crd_guard_test.go` strips `blockDeviceRefs` from the template; `virtualDiskTemplates` is the sole source, the controller builds the refs). Fixed in 375a56d: dropped the stale paragraph and aligned the RU `virtualDiskTemplates` description with the English CRD.

images/virtualization-artifact/pkg/controller/vmpool/internal/expectations/expectations.go
```diff
@@ -0,0 +1,172 @@
...
+	defer e.mu.Unlock()
+	it := e.getOrCreate(key)
+	it.creations += n
+	it.timestamp = e.now()
+}
+
+// ExpectDeletions records that the controller has just deleted the children
+// with the given UIDs for key and expects to observe their deletion events. It
+// resets the expectation's timestamp.
+func (e *Expectations) ExpectDeletions(key string, uids []types.UID) {
```
yaroslavborbat: uids ...types.UID

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
```diff
@@ -0,0 +1,336 @@
...
+
+// pickVictims chooses which replicas to remove during anonymous scale-down,
+// honouring the pool's scaleDownPolicy. Explicit forbids anonymous removal, so
+// it returns nothing — such pools shrink only through addressed removal.
+func pickVictims(policy v1alpha2.ScaleDownPolicy, candidates []v1alpha2.VirtualMachine, n int) []v1alpha2.VirtualMachine {
+	if n <= 0 || policy == v1alpha2.ScaleDownPolicyExplicit {
+		return nil
+	}
+	oldestFirst := policy == v1alpha2.ScaleDownPolicyOldestFirst
+	sort.SliceStable(candidates, func(i, j int) bool {
```
yaroslavborbat: use slices.SortStableFunc()

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
```diff
@@ -0,0 +1,336 @@
...
+		ti := candidates[i].GetCreationTimestamp().Time
```
yaroslavborbat: use cmp

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
```diff
@@ -0,0 +1,336 @@
...
+	if n > len(candidates) {
+		n = len(candidates)
+	}
+	return candidates[:n]
+}
+
+func (h *SyncHandler) newMember(pool *v1alpha2.VirtualMachinePool) *v1alpha2.VirtualMachine {
+	tmpl := pool.Spec.VirtualMachineTemplate
+
+	labels := make(map[string]string, len(tmpl.Metadata.Labels)+3)
```
yaroslavborbat, line 213: why +3 ? please drop a comment of fix it

images/virtualization-artifact/pkg/controller/vmpool/vmpool_scale_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/deckhouse/virtualization/api/core/v1alpha2"
+)
+
+// ScaleWebhookPath is where the scale-subresource guard is served. It must match
+// the ValidatingWebhookConfiguration entry for virtualmachinepools/scale.
+const ScaleWebhookPath = "/validate-virtualization-deckhouse-io-v1alpha2-virtualmachinepool-scale"
+
+// SetupScaleWebhook registers the guard that rejects anonymous scale-down via
+// the scale subresource for pools with scaleDownPolicy: Explicit.
+func SetupScaleWebhook(mgr manager.Manager) {
```
yaroslavborbat, line 44: need validate virtualmachine spec in other webhook. see vm controller

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
```diff
@@ -0,0 +1,124 @@
...
+			// template lists disk-template placeholders (kind VirtualDisk, name = a
+			// virtualDiskTemplates entry) plus any shared images; the disks handler
+			// has already resolved each placeholder to this member's concrete disk
+			// (e.g. "system" -> "web-a-system", or a reuse disk). Re-copying the
+			// template placeholders would dangle (no such disk) and duplicate the
+			// resolved ref. Disk add/remove is reconciled by the disks handler;
+			// blockDeviceRefs edits (reorder, added/removed image) reach a live
+			// replica on its next recreation (rotation/scale-up), like other
+			// disruptive template changes the pool does not force.
+			patched.Spec.BlockDeviceRefs = append([]v1alpha2.BlockDeviceSpecRef(nil), m.Spec.BlockDeviceRefs...)
```
yaroslavborbat: dirty hacks

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
```diff
@@ -0,0 +1,124 @@
...
+			if patched.Annotations == nil {
+				patched.Annotations = map[string]string{}
```
yaroslavborbat: make

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,670 @@
...
+		assignedThisPass[pick.Name] = true
+		logf.FromContext(ctx).Info("reusing a free pool disk", "member", m.GetName(), "disk", pick.Name, "diskTemplate", dt.Name)
+		return h.attachDisk(ctx, m, pick.Name, dt.Name)
+	}
+
+	// No free disk at all — create a new pool-owned disk and attach it.
+	name := fmt.Sprintf("%s-%s-%s", pool.GetName(), dt.Name, rand.String(6))
+	logf.FromContext(ctx).Info("creating a reuse disk", "member", m.GetName(), "disk", name, "diskTemplate", dt.Name)
+	if err := h.client.Create(ctx, h.newRetainDisk(pool, dt, name)); err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("create reuse disk %s: %w", name, err)
```
yaroslavborbat, line 527: return client.IgnoreAlreadyExists(err)

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,670 @@
...
+	}
+}
+
+// attachDisk makes the member reference diskName. If the member still carries an
+// unresolved placeholder ref (a blockDeviceRefs entry whose name equals the disk
+// template name, i.e. the user referenced the template by name), the placeholder
+// is replaced in place so the disk keeps its position in the boot order;
+// otherwise the ref is appended. Idempotent: a member already referencing
+// diskName is left untouched.
+func (h *DisksHandler) attachDisk(ctx context.Context, m *v1alpha2.VirtualMachine, diskName, placeholder string) error {
```
yaroslavborbat: m is not readable, replace with vm or member

yaroslavborbat: and other places too

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,670 @@
...
+	return dt.Reclaim.OnScaleDown == "" || dt.Reclaim.OnScaleDown == v1alpha2.VirtualDiskReclaimDelete
+}
+
+func (h *DisksHandler) ensureDeleteDisk(ctx context.Context, pool *v1alpha2.VirtualMachinePool, m *v1alpha2.VirtualMachine, dt v1alpha2.VirtualDiskTemplateSpec) error {
+	diskName := poollabels.DeleteDiskName(m.GetName(), dt.Name)
+
+	var disk v1alpha2.VirtualDisk
+	err := h.client.Get(ctx, types.NamespacedName{Namespace: m.GetNamespace(), Name: diskName}, &disk)
+	switch {
+	case apierrors.IsNotFound(err):
```
yaroslavborbat, line 617: check if err != nil first

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,670 @@
...
+const disksHandlerName = "disks"
+
+// DisksHandler reconciles per-replica disks. It is idempotent and self-healing:
+// for every live member it ensures each Delete-policy disk exists (owned by the
+// VirtualMachine, so it cascades away with the replica) and is referenced by the
+// member. Retain-policy (reusable) disks are handled by a later slice.
+type DisksHandler struct {
+	client client.Client
+	// now is injectable so tests can control free-disk ageing deterministically.
+	now func() time.Time
```
yaroslavborbat: why now is needed? maybe can be replaced with a clock?

images/virtualization-artifact/pkg/controller/vmpool/internal/watcher/vm.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return &types.NamespacedName{Namespace: vm.GetNamespace(), Name: ref.Name}
+}
+
+func (m *memberEventHandler) Create(_ context.Context, e event.TypedCreateEvent[*v1alpha2.VirtualMachine], q workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	key := ownerKey(e.Object)
+	if key == nil {
+		return
+	}
+	m.exp.CreationObserved(key.String())
```
yaroslavborbat, line 81: please write ut for expectations and watcher

images/virtualization-artifact/pkg/controller/vmpool/crd_guard_test.go
```diff
@@ -0,0 +1,106 @@
+/*
```
yaroslavborbat: is it needed?

test/e2e/vmpool/vmpool.go
```diff
@@ -0,0 +1,257 @@
...
+		f.Before()
+		DeferCleanup(f.After)
+	})
+
+	// buildPool returns a pool of tiny VMs with a single per-replica root disk
+	// template. The template's blockDeviceRefs references that disk by name (a
+	// placeholder the controller resolves per replica); the bijection with
+	// virtualDiskTemplates is enforced by admission.
+	buildPool := func(replicas int32, policy v1alpha2.ScaleDownPolicy, reclaim v1alpha2.VirtualDiskReclaim) *v1alpha2.VirtualMachinePool {
+		tmpl := vmbuilder.New(
```
yaroslavborbat: Why didn't anybody use NewMinimalVM?

test/e2e/vmpool/vmpool.go
```diff
@@ -0,0 +1,257 @@
...
+		return n
+	}
+
+	// scaleDownWith POSTs to the aggregated-apiserver scaleDownWith subresource,
+	// the same call `kubectl create --raw` makes. Returns the request error so
+	// both the success and the rejection paths can be asserted.
+	scaleDownWith := func(poolName string, targets ...string) error {
+		body, err := json.Marshal(map[string][]string{"targets": targets})
+		Expect(err).NotTo(HaveOccurred())
+		return f.KubeClient().Discovery().RESTClient().Post().
```
yaroslavborbat: add subresource to client and use it
see https://github.com/deckhouse/virtualization/blob/e132f73fc72c04ad8db3fc14144a16503bf7f4da/api/client/kubeclient/client.go#L53
https://github.com/deckhouse/virtualization/blob/e132f73fc72c04ad8db3fc14144a16503bf7f4da/api/client/kubeclient/vm.go#L53

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
```diff
@@ -0,0 +1,336 @@
...
+func (h *SyncHandler) listMembers(ctx context.Context, pool *v1alpha2.VirtualMachinePool) ([]v1alpha2.VirtualMachine, error) {
+	return poollabels.ListMembers(ctx, h.client, pool)
+}
+
+func (h *SyncHandler) scaleUp(ctx context.Context, pool *v1alpha2.VirtualMachinePool, key string, n int) error {
+	// Record the expectation before creating so a create event cannot be observed
+	// before we start waiting for it.
+	h.exp.ExpectCreations(key, n)
+	var errs error
+	for i := 0; i < n; i++ {
```
yaroslavborbat: ancient version

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
yaroslavborbat: and other places

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,670 @@
...
+	// ended up on two members, detach it from the stuck one so it is reassigned.
+	if err := h.reassignCollisions(ctx, pool, members); err != nil {
+		errs = errors.Join(errs, err)
+	}
+
+	// After (re)assignment, garbage-collect free reuse disks per Retain template.
+	// Track the soonest a free disk becomes GC-eligible and requeue for it, so ttl
+	// collection fires even when nothing else triggers a reconcile (idle pool).
+	var requeueAfter time.Duration
+	for j := range pool.Spec.VirtualDiskTemplates {
```
yaroslavborbat: j?

yaroslavborbat (changes requested):

images/virtualization-artifact/pkg/controller/vmpool/vmpool_scale_webhook.go
fl64: Agreed there's no up-front validation of the template's VM spec today — a bad `virtualMachineTemplate.spec` only surfaces as a `FailedCreate` event once the controller tries to create a member. Before wiring this up, could you confirm the intended scope?

1. Run the full `vm.NewValidator` chain against a VM synthesized from the template (it's clientful — we'd build it with the controller client + block-device/attachment services), or only the pure spec-level validators (block-device-refs, first-disk, cpu-count, …)? Some validators (IPAM, sizing-policy, PV node-affinity) assume a concrete, schedulable VM and may be meaningless for a template.
2. Validate on both CREATE and UPDATE of the pool?

I'll capture the decision in the ADR (MR !238) once we settle the depth.

images/virtualization-artifact/pkg/apiserver/registry/vmpool/storage/storage.go
fl64: Already addressed in 375a56d — `Get` now mirrors the VM storage: it returns a meta-object when the pool exists and `NewNotFound` only when it truly does not.

images/virtualization-artifact/pkg/controller/vmpool/internal/expectations/expectations.go
fl64: Done in 4c27140 — `ExpectDeletions` is now variadic.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
fl64: Done in 4c27140 — `pickVictims` and `gcReuseDisks` now use `slices.SortStableFunc`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
fl64: Done in 4c27140 — the comparator returns an int via `time.Time.Compare`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
fl64: Done in 4c27140 — added a comment: the +3 covers the managed labels (pool-uid, pool, template-hash).

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
fl64: Done in 72fea41 — extracted into a named `applyTemplateSpec` helper.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
fl64: Done in 72fea41 — uses `make(map[string]string)`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 4c27140 — uses `client.IgnoreAlreadyExists`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done — renamed `m` to `vm` across disks.go (4c27140) and template.go (72fea41).

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 4c27140 — the switch checks the non-NotFound error first.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 4c27140 — injects a `clock.PassiveClock` (`clock.RealClock` in prod, a fake clock in tests).

images/virtualization-artifact/pkg/controller/vmpool/internal/watcher/vm.go
fl64: Done in 74337bf — added watcher unit tests (`ownerKey` plus the create/delete/update handlers against the expectations tracker and the workqueue). Expectations already had its own suite.

images/virtualization-artifact/pkg/controller/vmpool/crd_guard_test.go
fl64: Removed in 72fea41 — it only asserted the generated CRD shape, not that the CEL rule actually works. The blockDeviceRefs/virtualDiskTemplates bijection is now covered by admission-rejection cases in the e2e suite (74337bf).

test/e2e/vmpool/vmpool.go
fl64: Done in 74337bf — the pool template is now built with `object.NewMinimalVM`.

test/e2e/vmpool/vmpool.go
fl64: Done — added a typed `VirtualMachinePools().ScaleDownWith()` client (00b39ab); the e2e now uses it (74337bf).

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/sync.go
fl64: Done in 4c27140 — replaced with a range-over-int loop.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 4c27140 — renamed to `i`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
```diff
@@ -0,0 +1,130 @@
...
+// applyTemplateSpec returns a copy of vm with its spec brought to the pool
+// template's spec while keeping the member's own blockDeviceRefs. The template
+// lists disk-template placeholders (kind VirtualDisk, name = a virtualDiskTemplates
+// entry) plus any shared images; the disks handler has already resolved each
+// placeholder to this member's concrete disk (e.g. "system" -> "web-a-system", or a
+// reuse disk). Re-copying the template placeholders would dangle (no such disk) and
+// duplicate the resolved ref. Disk add/remove is reconciled by the disks handler;
+// blockDeviceRefs edits (reorder, added/removed image) reach a live replica on its
+// next recreation (rotation/scale-up), like other disruptive template changes the
+// pool does not force.
```
yaroslavborbat: comment is longer than code

yaroslavborbat: i cannot understand both

images/virtualization-artifact/cmd/virtualization-controller/main.go
```diff
@@ -487,6 +488,19 @@ func main() {
...
+		log.Error(err.Error())
+		os.Exit(1)
+	}
+	// Guards anonymous scale-down for scaleDownPolicy: Explicit; self-gated by the VirtualMachinePool feature gate.
+	vmpool.SetupScaleWebhook(mgr)
+	// Validates the embedded virtualMachineTemplate spec on pool create/update; self-gated by the feature gate.
+	if err = vmpool.SetupValidationWebhook(mgr, vmpoolLogger); err != nil {
+		log.Error(err.Error())
+		os.Exit(1)
+	}
```
yaroslavborbat: usually, we setup webhooks in SetupController

yaroslavborbat (changes requested):

yaroslavborbat:
impact level low???

images/virtualization-artifact/pkg/controller/vmpool/vmpool_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+		Complete()
+}
+
+// poolValidator validates a VirtualMachinePool by running the template-spec
+// validators against a VirtualMachine synthesized from the pool template.
+type poolValidator struct {
+	vmValidator *vm.Validator
+}
+
+func (v *poolValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
yaroslavborbat, line 86: Vm name must be less than 63 characters. Now, pool vms generated includes pool name. Pool name should be less than 57 characters. 57?

yaroslavborbat, line 86: include disk validation too

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,671 @@
...
+	// resourceVersion) instead of clobbering them from the stale original.
+	updated.DeepCopyInto(vm)
+	return nil
+}
+
+func isDeletePolicy(dt v1alpha2.VirtualDiskTemplateSpec) bool {
+	return dt.Reclaim.OnScaleDown == "" || dt.Reclaim.OnScaleDown == v1alpha2.VirtualDiskReclaimDelete
+}
+
+func (h *DisksHandler) ensureDeleteDisk(ctx context.Context, pool *v1alpha2.VirtualMachinePool, vm *v1alpha2.VirtualMachine, dt v1alpha2.VirtualDiskTemplateSpec) error {
```
yaroslavborbat: diskTemplate

yaroslavborbat (changes requested):

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,671 @@
...
+	diskName := poollabels.DeleteDiskName(vm.GetName(), dt.Name)
+
+	var disk v1alpha2.VirtualDisk
+	err := h.client.Get(ctx, types.NamespacedName{Namespace: vm.GetNamespace(), Name: diskName}, &disk)
+	switch {
+	case err != nil && !apierrors.IsNotFound(err):
```
yaroslavborbat: switch {
case err == nil: // do nothing
case apierrors.IsNotFound(err):
  ...
default:
  ...
}

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,671 @@
...
+		}
+		if len(refs) == len(cur.Spec.BlockDeviceRefs) {
+			cur.DeepCopyInto(vm) // already detached; sync the caller's copy
+			return nil
+		}
+		cur.Spec.BlockDeviceRefs = refs
+		if err := h.client.Update(ctx, cur); err != nil {
+			return err
+		}
+		cur.DeepCopyInto(vm)
```
yaroslavborbat: no needed

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
```diff
@@ -0,0 +1,671 @@
...
+		}
+		refs := make([]v1alpha2.BlockDeviceSpecRef, 0, len(cur.Spec.BlockDeviceRefs))
+		for _, ref := range cur.Spec.BlockDeviceRefs {
+			if ref.Kind == v1alpha2.DiskDevice && ref.Name == diskName {
+				continue
+			}
+			refs = append(refs, ref)
+		}
+		if len(refs) == len(cur.Spec.BlockDeviceRefs) {
+			cur.DeepCopyInto(vm) // already detached; sync the caller's copy
```
yaroslavborbat: no need  
what is a problem it solved?

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		// Step 2: the spec is on the desired revision. Mark the revision label as
+		// effectively applied only once the disruptive part is no longer pending;
+		// while the VM awaits a restart the label stays on the old revision.
+		if awaitingRestart(vm) {
+			continue
+		}
+		if vm.GetLabels()[poollabels.TemplateHash] != desiredHash {
+			updated := vm.DeepCopy()
```
yaroslavborbat: no need

yaroslavborbat: please check other execution DeepCopy methods in pr. you get and modify vm in one func, you shouldn't  worryyy about something modify this obj too

yaroslavborbat (changes requested):

images/virtualization-artifact/cmd/virtualization-controller/main.go
fl64: Done in ab6d1e6b — both the scale and validation webhooks are now set up from SetupController.

images/virtualization-artifact/pkg/controller/vmpool/vmpool_webhook.go
fl64: Done in ab6d1e6b — the webhook rejects a pool name longer than 57 chars, so `<pool>-<rand>` member names stay under 63.

fl64: Done in ab6d1e6b — each `virtualDiskTemplates[].spec` is now validated (PVC size, ISO source) via `vd.NewTemplateSpecValidator`. Storage-class/migration checks are left out (need a live disk).

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 0ef912dd — renamed `dt` to `diskTemplate`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Done in 0ef912dd — `switch { case err == nil / IsNotFound / default }`.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Removed in 0ef912dd.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/disks.go
fl64: Removed in 0ef912dd — dropped the short-circuit; detach just filters the refs and updates.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
fl64: Simplified in 0ef912dd — `applyTemplateSpec` mutates the member in place (no defensive DeepCopy) and the comment is shorter.

images/virtualization-artifact/pkg/controller/vmpool/internal/handler/template.go
fl64: Done in 0ef912dd — no more defensive DeepCopy; the member is modified in place within the reconcile.

images/virtualization-artifact/pkg/controller/vmpool/vmpool_scale_webhook.go
fl64: Sorry for the wall of text. Short version — the pool webhook now validates the template up front (ab6d1e6b):

- **VM spec**: block-device-refs, first-disk, cpu-count, sizing-policy, networks, affinity, topology-spread, meta.
- **Disk templates**: PVC size + ISO source.
- **Pool name**: rejected if > 57 chars (so `<pool>-<rand>` fits in 63).

Left out on purpose: IPAM, block-device-limiter, PV-node-affinity — they need a real VM/disk that doesn't exist at pool admission, and every real replica still goes through the full VM + VirtualDisk webhooks when the controller creates it. WDYT?

api/core/v1alpha2/virtual_machine_pool.go
```diff
@@ -0,0 +1,265 @@
...
+// virtualDiskTemplates describes each per-replica disk, and the template's
+// blockDeviceRefs references those disks (by name, kind VirtualDisk) to set the
+// boot order and interleave shared images. The rule below enforces a bijection —
+// every template is referenced exactly once, and every VirtualDisk reference
+// resolves to a template — so neither list can carry a dangling entry.
+//
+// +kubebuilder:validation:XValidation:rule="has(self.virtualMachineTemplate.spec) && has(self.virtualMachineTemplate.spec.blockDeviceRefs) && self.virtualDiskTemplates.all(t, self.virtualMachineTemplate.spec.blockDeviceRefs.exists(r, r.kind == 'VirtualDisk' && r.name == t.name)) && self.virtualMachineTemplate.spec.blockDeviceRefs.filter(r, r.kind == 'VirtualDisk').all(r, self.virtualDiskTemplates.exists(t, t.name == r.name)) && self.virtualMachineTemplate.spec.blockDeviceRefs.filter(r, r.kind == 'VirtualDisk').size() == self.virtualDiskTemplates.size()",message="each virtualDiskTemplates entry must be referenced exactly once by a VirtualDisk entry in virtualMachineTemplate.spec.blockDeviceRefs, and every VirtualDisk reference must name a virtualDiskTemplates entry"
```
yaroslavborbat: difficult for understand. please separate to several comments

yaroslavborbat: if possible

yaroslavborbat (changes requested):

yaroslavborbat (dismissed):

fl64:
Fixed — dropped the `impact_level` line (this repo has only `low`/`high`, no `default`), so the entry now goes into the changelog.

api/core/v1alpha2/virtual_machine_pool.go
fl64: Done in 9645a20 — split into three separate `XValidation` rules, each with its own message: (1) every `virtualDiskTemplates` entry is referenced by a VirtualDisk, (2) every VirtualDisk reference names a template, (3) one-to-one (no duplicate references).

yaroslavborbat (approved):

PR #2586, opened 2026-07-03
feat(dvcr): add per-namespace authorization

danilrwx:
## Description

This PR upgrades DVCR (the Deckhouse Virtualization container registry) and adds optional per-namespace authorization for it.

**DVCR registry upgrade (behavior-preserving):**
- Upgrade the embedded `distribution` registry from 2.8.3 to 3.1.1, migrating the build from GOPATH to a Go module (`github.com/distribution/distribution/v3`). Config schema, on-disk storage layout, `dvcr-cleaner` and garbage collection are unchanged.

**Per-namespace DVCR authorization (new, behind `dvcr.tenantRegistryAuthorization`, default `false`):**
- A new in-registry `dvcr-k8s` `AccessController` is compiled into the registry binary. It authorizes by **scoped token**, not by caller identity: the controller mints a short-lived JWT granting access to exactly the repositories a given Pod uses, and the registry verifies the signature (reusing distribution's own `registry/auth/token` backend) and allows only the operations listed in the token. Fail-closed.
- The token format is distribution's native JWT (`access` claim), so verification reuses upstream code; `go-jose` is already in distribution's dependency graph.
- The signing key is an asymmetric ES256 keypair generated into `dvcr-secrets` (private → controller, public → registry). A compromised registry cannot mint tokens.
- importer/uploader Pods and CDI DataVolumes authenticate to DVCR with the scoped token presented as ordinary Basic credentials. The shared read-write credential is no longer copied into tenant namespaces, and no CDI fork or `TokenReview` is required.
- Nodes (containerd) use a dedicated pull-only `node-puller` credential; the controller keeps the static admin credential.
- When the flag is off, rendered manifests and behavior are identical to today (htpasswd).

DVCR is a single cluster-wide registry shared by all namespaces, and image import/upload previously relied on one shared read-write credential (copied into tenant namespaces) with no per-repository authorization. A tenant able to read Secrets in its own namespace could exfiltrate that credential and read or overwrite any other namespace's images. This change adds optional multi-tenant isolation so a Pod can access only the repositories it legitimately needs.

Authorizing by a signed scoped token (rather than by the caller's namespace) is what makes this correct even for imports whose Pod must run in another namespace to mount a source PVC (e.g. a ClusterVirtualImage built from a VirtualDisk): the token names the exact repository, independent of where the Pod runs.

The v3 upgrade is a prerequisite: the Go-module build is what allows compiling the in-registry authorization backend and reusing its token verification.

- **Flag off (default):** identical to current behavior (htpasswd) — verified with `helm template` for both states.
- **Flag on:** each import/upload Pod gets a token scoped to its own repositories; cross-repository access is denied (403); no read-write DVCR credential Secret is copied into tenant namespaces; nodes still pull VM images (pull-only); the controller still operates.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: module
type: feature
summary: Optional per-namespace authorization for DVCR (`dvcr.tenantRegistryAuthorization`) isolating container image access between namespaces.
```


images/hooks/pkg/hooks/generate-secret-for-dvcr/hook.go
```diff
@@ -150,10 +201,72 @@ func getDVCRSecretsFromSecrets(input *pkg.HookInput) (dvcrSecretData, error) {
 
 func getDVCRSecretsFromValues(input *pkg.HookInput) dvcrSecretData {
 	return dvcrSecretData{
-		PasswordRW: input.Values.Get(passwordRWValuePath).String(),
-		Salt:       input.Values.Get(saltValuePath).String(),
-		Htpasswd:   input.Values.Get(htpasswdValuePath).String(),
+		PasswordRW:      input.Values.Get(passwordRWValuePath).String(),
+		PasswordPuller:  input.Values.Get(passwordPullerValuePath).String(),
```
yaroslavborbat: PasswordRO will be better. in one style

openapi/config-values.yaml
```diff
@@ -133,6 +133,18 @@ properties:
               Schedule to run garbage collection procedure that remove stale images for `ClusterVirtualImage`, `VirtualImage`, `VirtualDisk` resources deleted from the cluster.
 
               By default, periodic garbage collection is enabled and runs daily at 02:00.
+      tenantRegistryAuthorization:
```
yaroslavborbat: why is it needed?

images/virtualization-artifact/pkg/dvcr/registrytoken/signer_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Errorf("aud = %v, want %q", claims["aud"], Audience)
+	}
+	acc, ok := claims["access"].([]interface{})
+	if !ok || len(acc) != 1 {
+		t.Fatalf("access claim = %v", claims["access"])
+	}
+	entry := acc[0].(map[string]interface{})
+	if entry["name"] != "cvi/my-image" || entry["type"] != "repository" {
+		t.Errorf("access entry = %v", entry)
+	}
```
yaroslavborbat: use require please or ginkgo with goomega

images/dvcr/dvcr-k8s-auth/access_test.go
```diff
@@ -0,0 +1,220 @@
...
+func TestVerifyJWT_Rejections(t *testing.T) {
+	key := newKey(t)
+	ac := newController(t, &key.PublicKey)
+	now := time.Now()
+
+	t.Run("expired", func(t *testing.T) {
+		claims := testClaims(now)
+		claims["exp"] = now.Add(-time.Hour).Unix()
+		if _, err := ac.verifyJWT(mintKidToken(t, key, claims)); err == nil {
+			t.Fatal("expired token accepted")
```
yaroslavborbat: require

yaroslavborbat (changes requested):

yaroslavborbat (approved):

PR #2614, opened 2026-07-08
fix(vm): secureboot vm no default storageclass

danilrwx:
## Description

A SecureBoot virtual machine stores its Secure Boot state in a persistent volume, which needs a default StorageClass. On a cluster without one the volume can't be created and the VM never starts — and nothing on the VirtualMachine explained why, the error was only visible on internal objects.

Now the VirtualMachine reports a clear reason in this case instead of silently staying pending. Start-failure messages are also cleaned up to plain user-facing wording (no internal pod phases or reason codes), and the default StorageClass requirement is documented next to the bootloader settings.

Enabling SecureBoot on a cluster without a default StorageClass looked like a hang: the VM stayed pending with no visible cause. Now the reason is shown on the VirtualMachine itself and the requirement is spelled out in the docs.

- No default StorageClass in the cluster → the SecureBoot VM's `Running` condition becomes `False` with a message like `Cannot provision storage for the virtual machine's Secure Boot state: no default storage class found.`, and the VM starts automatically once a default StorageClass is set.
- Other start failures (image pull, scheduling, disk provisioning) show plain user-facing messages instead of raw internal statuses.
- Non-SecureBoot VMs are unaffected.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: "A SecureBoot virtual machine now reports on the VirtualMachine why it cannot start (for example, a missing default StorageClass) instead of staying pending without explanation."
```


docs/USER_GUIDE.ru.md
```diff
@@ -1439,6 +1428,10 @@ spec:
 Для большинства современных Linux-дистрибутивов рекомендуется использовать `bootloader: EFI`. Для Windows обычно используют `bootloader: EFI` или `bootloader: EFIWithSecureBoot`.
 {{< /alert >}}
 
+{{< alert level="warning" >}}
+`EFIWithSecureBoot` хранит состояние Secure Boot в постоянном томе, поэтому в кластере должен быть StorageClass по умолчанию. Если его нет, виртуальная машина остаётся в состоянии `Pending` (в условии `Running` отображается сообщение `no default storage class found`), пока StorageClass по умолчанию не будет задан.
```
z9r5: > в условии `Running`

Тут бы переформулировать может. Не понятно кмк.

danilrwx: Reworded it: dropped the reference to the `Running` condition and the raw message. It now describes the effect in user terms — the VM does not start and stays in `Pending`, and its status reports that no default StorageClass is found; once a default StorageClass exists, the VM starts automatically. Updated in both the RU and EN guides.

Isteb4k (approved):

Isteb4k (approved):

PR #2622, opened 2026-07-09
chore(api): integrate crd-enricher into CRD codegen

prismagod:
## Description
Integrate Deckhouse `crd-enricher` into the API CRD generation pipeline.
After `controller-gen`, enricher runs on the temporary CRD output (before files are copied to `crds/`), matching the flow in [deckhouse#21220](https://github.com/deckhouse/deckhouse/pull/21220).

This enables `+crd-enricher:raw:*` and other markers in Go API types without editing generated CRD YAML by hand.

- `cd api && task generate:crds` completes successfully (Docker required for `format:yaml`)
- With no enricher markers, generated CRDs are unchanged after prettier

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: api
type: chore
summary: "Integrate crd-enricher into the API CRD codegen pipeline."
```


yaroslavborbat (dismissed):

crds/virtualmachines.yaml
```diff
@@ -35,6 +35,24 @@ spec:
             - `.spec.runPolicy`.
 
             **Note:** The `metadata.name` field must comply with [Kubernetes object naming conventions](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/) and must not exceed 63 characters.
+          x-doc-examples:
+            - apiVersion: virtualization.deckhouse.io/v1alpha2
+              kind: VirtualMachine
+              metadata:
+                name: example-vm
+              spec:
+                virtualMachineClassName: default
```
diafour: `default` is ambiguous here.

```suggestion
                virtualMachineClassName: cpu-host-on-worker-nodes
```

crds/virtualmachines.yaml
```diff
@@ -35,6 +35,24 @@ spec:
...
+                cpu:
+                  cores: 1
+                  coreFraction: 10%
+                memory:
+                  size: 1Gi
+                provisioning:
+                  type: UserData
+                  userData: <USER_DATA>
```
diafour:
```suggestion
                  type: UserDataRef
                  userDataRef:
                    kind: Secret
                    name: cloudinit-for-vms
```

crds/virtualmachines.yaml
prismagod: Thanks, applied to both .go and .yaml files.

crds/virtualmachines.yaml
prismagod: Thanks, applied to both .go and .yaml files.

diafour (dismissed):

diafour (approved):

PR #2632, opened 2026-07-10
fix(observability): correct audit server default TLS paths and prefix resources with d8-

danilrwx:
## Description

Two fixes in the audit delivery path (log-shipper → virtualization-audit).

The audit server crashed on startup when launched without explicit TLS flags: its built-in default certificate paths pointed at a directory that does not exist, while the mounted secret lives under a different path. The defaults now match the real mount location, so the server loads TLS and starts even with no flags passed. In the shipped deployment the paths are passed explicitly, so this was a latent footgun (it bites debug/alternative launches).

Separately, the cluster-scoped log-shipper objects created for audit delivery are renamed to follow the module's `d8-` naming convention for cluster-scoped resources (the old `to-`/unprefixed names are replaced).

The default certificate paths were a latent trap — any launch without the explicit flags fails on missing TLS files, which is confusing to debug. And the audit log-shipper resource names did not follow the `d8-` convention used for module-owned cluster-scoped objects, unlike the rest of the module.

- The audit server starts and loads its certificate, key and CA from the mounted certificates directory even without explicit `--tls-*` flags.
- The audit log-shipper resources are named `d8-virtualization-audit*`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: observability
type: fix
summary: "The audit server no longer fails to start when TLS certificate paths are not passed explicitly."
---
section: observability
type: chore
summary: "Audit log-shipper resources are renamed to the d8- prefixed naming convention."
impact_level: low
```


universal-itengineer (approved):

deckhouse-BOaTswain:
Cherry pick PR [2637](https://github.com/deckhouse/virtualization/pull/2637) to the branch [release-1.9](https://github.com/deckhouse/virtualization/tree/release-1.9) successful!

PR #2638, opened 2026-07-10
fix(dra): drop redundant /sys hostPath mount breaking container start

universal-itengineer:
## Description

The `virtualization-dra` USB DaemonSet pods never start — they stay in `CrashLoopBackOff` on every node, so USB device passthrough is unavailable. The main container is privileged and runs with a read-only root filesystem while also bind-mounting the host `/sys`. Under containerd this combination makes the runtime fail to set up the `/sys/fs/cgroup` mount on the read-only rootfs, and the container never starts.

A privileged container already gets a read-write `/sys` from the runtime, so the explicit `/sys` mount is redundant. It is removed together with its volume and the now-dead `/sys` entry in the SecurityPolicy host-path allow-list.

Without this, the DRA component cannot run at all, and USB devices cannot be passed through to VMs. Verified on a cluster: a privileged pod with a read-only rootfs and an explicit `/sys` mount crash-loops with the cgroup mount error, while the same pod without that mount starts fine and still has read-write access to `/sys/bus/usb`.

`virtualization-dra` pods reach `Running` on all nodes (no `CrashLoopBackOff`), and USB device passthrough works.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section: vm
type: fix
summary: "USB device passthrough pods no longer fail to start on nodes with a read-only container root filesystem."
```

### DIFF
--- a/templates/virtualization-dra/daemonset.yaml
+++ b/templates/virtualization-dra/daemonset.yaml
@@ -154,11 +154,6 @@ spec:
           metadata:
             description: |
               Allow read-write access to /var/run hostPath volume.
-        - path: /sys
-          readOnly: false
-          metadata:
-            description: |
-              Allow read-write access to /sys hostPath volume.
 {{- end }}
 
 ---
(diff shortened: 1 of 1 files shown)


diafour (approved):

deckhouse-BOaTswain:
Cherry pick PR [2661](https://github.com/deckhouse/virtualization/pull/2661) to the branch [release-1.9](https://github.com/deckhouse/virtualization/tree/release-1.9) successful!

PR #2645, opened 2026-07-13
fix(vm): preserve scsi bus for VMBDA disks when enableParavirtualization=false

universal-itengineer:
## Description
On a VM with `enableParavirtualization=false`, disks attached via VirtualMachineBlockDeviceAttachment were switched to the SATA bus, even though hot-plugged disks must stay on SCSI. This broke such attachments. The bus of a runtime hot-plugged disk is now preserved regardless of the paravirtualization mode.

Regression from #2624: that change moves a device to the bus implied by the paravirtualization mode when the mode flips. But a VMBDA disk is always hot-plugged on SCSI independently of that mode, so on a non-paravirtualized VM it was wrongly rewritten to SATA — an invalid bus for a hot-plugged device. Static disks declared in the VM spec still follow the paravirtualization bus on the restart they already require.

Create a VM with `enableParavirtualization: false`, then attach a VirtualDisk via VirtualMachineBlockDeviceAttachment. The disk stays on the SCSI bus and attaches successfully.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: vm
type: fix
summary: "Block devices attached to a VM with disabled paravirtualization are no longer switched to the SATA bus and attach correctly."
```

### DIFF
--- a/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
+++ b/images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
@@ -516,45 +516,18 @@ func (b *KVVM) ClearDisks() {
 	b.Resource.Spec.Template.Spec.Volumes = nil
 }
 
-func (b *KVVM) getExistingDiskBus(name string) virtv1.DiskBus {
-	for _, d := range b.Resource.Spec.Template.Spec.Domain.Devices.Disks {
-		if d.Name != name {
-			continue
-		}
-		if d.CDRom != nil {
-			return d.CDRom.Bus
-		}
-		if d.Disk != nil {
-			return d.Disk.Bus
-		}
-	}
-	return ""
-}
-
 func (b *KVVM) SetDisk(name string, opts SetDiskOptions) error {
-	devPreset := DeviceOptionsPresets.Find(b.opts.EnableParavirtualization)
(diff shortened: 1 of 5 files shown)


images/virtualization-artifact/pkg/controller/kvbuilder/kvvm.go
```diff
@@ -516,48 +516,32 @@ func (b *KVVM) ClearDisks() {
...
-
 func (b *KVVM) SetDisk(name string, opts SetDiskOptions) error {
 	devPreset := DeviceOptionsPresets.Find(b.opts.EnableParavirtualization)
 
+	diskBus, cdromBus := devPreset.DiskBus, devPreset.CdromBus
+	// Hot-plugged disks are attached via AddVolume, which always uses the scsi bus
+	// regardless of the paravirtualization preset. Keep them on scsi so a VM with
+	// enableParavirtualization=false (sata preset) does not rewrite an already
+	// attached device to sata, which is invalid for a hot-plugged disk. Static
+	// disks follow the preset and move buses on the restart a paravirtualization
```
diafour:
```suggestion
	// disks follow the preset and change buses on the restart a paravirtualization
```

universal-itengineer: Done

diafour (approved):

PR #2647, opened 2026-07-13
fix(dra): load xz-compressed usbip kernel modules

danilrwx:
## Description

On nodes where the pre-installed usbip kernel modules are shipped compressed, the DRA USB driver failed to initialize with `module ...usbip-core.ko.zst not found`, so USB passthrough was unavailable on those nodes. Kernel modules can be compressed with any of the three formats the kernel supports (zstd, xz, gzip), but only `.ko` and `.ko.zst` were recognized. The driver now finds and decompresses `.ko.xz` and `.ko.gz` too, and the node-configuration check that installs the modules recognizes the same set of formats.

Some distributions (e.g. the observed node shipped `.ko.xz`) compress kernel modules with xz or gzip instead of zstd. USB passthrough silently stayed broken on such nodes because the driver only knew about zstd. This makes module loading work regardless of the host's chosen compression format.

On a node whose usbip modules are `.ko.xz` (or `.ko.gz`), the DRA USB driver init loads the modules successfully and USB passthrough works, instead of failing with `module not found`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries

```changes
section:  core
type: fix
summary: "USB passthrough now works on nodes where usbip kernel modules are compressed with xz or gzip, not only zstd."
impact_level: low
```

### DIFF
--- a/images/virtualization-dra/pkg/modprobe/modprobe.go
+++ b/images/virtualization-dra/pkg/modprobe/modprobe.go
@@ -17,6 +17,7 @@ limitations under the License.
 package modprobe
 
 import (
+	"compress/gzip"
 	"errors"
 	"fmt"
 	"io"
@@ -26,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/klauspost/compress/zstd"
+	"github.com/ulikunitz/xz"
 	"golang.org/x/sys/unix"
 )
 
@@ -40,7 +42,7 @@ func LoadModules(modules ...string) error {
 }
 
 func loadModule(path string) error {
-	if strings.HasSuffix(path, ".zst") {
+	if isCompressed(path) {
 		uncompressedPath, err := uncompressModuleToTmp(path)
 		if err != nil {
 			return fmt.Errorf("uncompress module %s: %w", path, err)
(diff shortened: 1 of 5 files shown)


images/virtualization-dra/pkg/modprobe/modprobe.go
```diff
@@ -99,6 +103,25 @@ func uncompressModuleToTmp(path string) (string, error) {
...
+func newDecompressor(path string, in io.Reader) (io.Reader, error) {
+	switch {
+	case strings.HasSuffix(path, ".xz"):
+		return xz.NewReader(in)
+	case strings.HasSuffix(path, ".gz"):
+		return gzip.NewReader(in)
+	default:
+		return zstd.NewReader(in)
+	}
+}
```
yaroslavborbat, line 125: add swith case with HasSuffix zst and default error

yaroslavborbat (changes requested):

images/virtualization-dra/pkg/modprobe/modprobe.go
danilrwx: Done - added an explicit `.zst` case and a `default` that returns an error for unsupported formats.

yaroslavborbat (approved):

PR #2677, opened 2026-07-14
fix(cli): keep collect-debug-info going when secondary resources are absent

LopatinDmitr:
## Description
`collect-debug-info` aborted with no archive at all when a secondary resource was missing — most often the KubeVirt VMI, which does not exist while a VM is stopped. Missing resources (and ones the caller has no access to) are now skipped with a warning and the rest of the bundle is still produced. A missing target VirtualMachine still fails, now with a clear message instead of a raw internal one.

Debug bundles are collected precisely when something is wrong — often on stopped or half-provisioned VMs. Aborting the whole command in that case left users with nothing to attach to a bug report.

`d8v collect-debug-info <vm> -n <ns> > debug.tar.gz` on a stopped VM exits 0 and the archive contains the VM config, events and available block devices; skipped resources are reported as warnings on stderr. An unknown name returns `VirtualMachine "<vm>" not found in namespace "<ns>"`.

## Checklist
- [ ] e2e tests passed.

## Changelog entries
```changes
section: cli
type: fix
summary: "collect-debug-info no longer fails to produce an archive when the VM is stopped or a referenced resource is missing."
```

### DIFF
--- a/src/cli/internal/cmd/collectdebuginfo/collectdebuginfo_test.go
+++ b/src/cli/internal/cmd/collectdebuginfo/collectdebuginfo_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 Flant JSC
+Copyright 2026 Flant JSC
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/cli/internal/cmd/collectdebuginfo/collectdebuginfo_test.go
```diff
@@ -0,0 +1,308 @@
+/*
+Copyright 2025 Flant JSC
```
diafour:
```suggestion
Copyright 2026 Flant JSC
```

diafour (approved):

